PR #20, opened 2020-06-08
Add and edit functionality for Tag

rathorevaibhav:
## Summary

This completes the create & edit functionality for tags using ApolloGraphQL client.

## Test Plan

To be done separately.

### DIFF
--- a/src/containers/Tag/Tag.tsx
+++ b/src/containers/Tag/Tag.tsx
@@ -132,7 +132,7 @@ export const Tag: React.SFC<TagProps> = (props) => {
   if (error) return <p>Error :(</p>;
 
   const saveHandler = () => {
-    const payload = {
+    const payload: tagTypes.Tag = {
       label: label,
       description: description,
       isActive: isActive,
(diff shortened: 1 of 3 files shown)


src/containers/Tag/Tag.tsx
```diff
@@ -1,58 +1,159 @@
...
-      id: Number(tag ? tagId : Math.floor(Math.random() * Math.floor(100))),
-      label: name,
+    const payload = {
+      label: label,
       description: description,
-      is_active: isActive,
-      is_reserved: isReserved,
-      language_id: Number(languageId),
-      parent_id: Number(parentId),
+      isActive: isActive,
```
kurund, line 138: Any specific reason for changing the keys and removing the type definition for this?

rathorevaibhav: Yeah. Couple of reasons:
1. The GraphQL requires camelCased keys in the payload.
2. in the type definition, we had id, which we don't need to pass to the API. It'll be set automatically.

kurund: I think it's a good practice to keep types. We can do the following:
* Update types to reflect these changes.
* Remove id

rathorevaibhav: Makes sense. Working on it.

rathorevaibhav: This is complete. Ready for review again.

Issue #14, opened 2020-06-08
Add unit tests for tag functionality


kurund:
**Describe the task**
* add test cases for tag functionality ( add / edit / delete )

**Expected behavior**
* All the test cases should pass

PR #25, opened 2020-06-09
Feature/formik-material

rathorevaibhav:
## Summary
Formik and Formik-material implementation as an alternative of Formsy.

## Test Plan

To be done separately in #14


kurund:
@rathorevaibhav Can you remove yarn.lock from your commit. I see you had added it .gitgnore

rathorevaibhav:
@kurund done with changes. Please check

rathorevaibhav:
Yeah, I had to run `yarn install` for the Formik packages and that updated yarn.lock

That is why it contains changes regarding the new packages

kurund:
@rathorevaibhav I think what you have done is correct. i.e adding it to .gitignore.
so it should not be listed as a part of your commit.

PR #29, opened 2020-06-11
Bye bye redux

kurund:
## Summary

Removed redux in favor of apolloclient for local state management

## Test Plan

Tag functionality is working fine.


src/containers/Tag/Tag.tsx
```diff
@@ -130,8 +129,8 @@ export const Tag: React.SFC<TagProps> = (props) => {
   if (loading) return <p>Loading...</p>;
   if (error) return <p>Error :(</p>;
 
-  const saveHandler = (tag: tagTypes.Tag) => {
```
rathorevaibhav, line 133: Checking on your thoughts on this. So we don't need to do type-hinting Tag anymore? Assuming this will entirely based on the tag structure at the backend and that is why we switched to `any`. Let me know if you had any additional thoughts.

kurund: We should do type checking. This is a temporary fix as we are removing redux. Just think of best place to add types. May be in grapql directory. We will need to grep for 'any' and the define specific types wherever possible.

rathorevaibhav: Sounds good. Let's merge then.

Issue #16, opened 2020-06-08
Implement snackbar component to display various messages in the application


kurund:
**Describe the task**
* Implement generic material UI snackbar component that could be used for various actions such as display success message, failure message, etc.
* Currently, implement this for tags
* Add test case for this functionality

**Expected behavior**
* We should display snackbar with the message after the appropriate action

**References**
https://material-ui.com/components/snackbars/

PR #34, opened 2020-06-12
Feature/notification alert

rathorevaibhav:
## Summary

Basic workflow of storing a message in Apollo cache during add/edit/delete tag and using JS alert method to show it in the Taglist page. The alert would be replaced by Snackbar in #16 

## Test Plan

To be covered in Snackbar PR #16

### DIFF
--- a/src/containers/Tag/TagList/TagList.tsx
+++ b/src/containers/Tag/TagList/TagList.tsx
@@ -1,8 +1,8 @@
 import React, { useState, useEffect } from 'react';
 import { Redirect, Link } from 'react-router-dom';
-import { useQuery, gql, useMutation } from '@apollo/client';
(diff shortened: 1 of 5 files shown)


src/containers/Tag/TagList/TagList.tsx
```diff
@@ -40,6 +45,14 @@ export const TagList: React.SFC<TagListProps> = (props) => {
     },
   });
 
+  useEffect(() => {
+    if (message.data && message.data.message) alert(message.data.message);
+    client.writeQuery({
```
kurund, line 50: Looks like this is the main code to write to local state and we are using the same in multiple places. It might make sense to make a common method that accepts what needed like message, type of message ( success, error etc)

Issue #13, opened 2020-06-08
Add pagination component


kurund:
**Describe the task**
* Shortlist pagination component and share it with the team
* Implement a pagination component for tags
* Abstract the same so that it could be easily used for other listings
* Add test case for this functionality
* Add test case for the pagination component

**Expected behavior**
* Pagination fetch the data from the backend.
* Implement filtering and sorting

**References**
Few options to evaluate:
* https://www.npmjs.com/package/react-data-table-component
* https://material-ui.com/components/pagination/

PR #36, opened 2020-06-13
Calvin/pagination

claalmve:
## Summary

This feature is for adding server-side Pagination to the `Tags` data table so that the user won't need to render potentially thousands of different objects on their end. It'll help with scalability over time and only requires a few more queries to be sent to the backend to work properly.

Addresses issue #13 

## Test Plan

I'll be generalizing the component and adding test cases for it and `TagList` soon.


src/index.tsx
```diff
@@ -12,7 +12,8 @@ import App from './App';
 import * as serviceWorker from './serviceWorker';
 
 ReactDOM.render(
-  <React.StrictMode>
+  // <React.StrictMode>
```
claalmve, line 15: I noticed before in a different project that React.StrictMode allows for render to happen twice in development for some reason. However, I wasn't totally sure if that behavior was consistent here too, so I can undo these deletions if need be.

src/containers/Tag/TagList/TagList.tsx
```diff
@@ -44,66 +95,147 @@ export const TagList: React.SFC<TagListProps> = (props) => {
...
-            </TableRow>
-          </TableHead>
-          <TableBody>{listing}</TableBody>
-        </Table>
-      </TableContainer>
-    </div>
+
+      {/* Table */}
+      {tagList ? (
+        <DataTable
```
claalmve, line 217: Will generalize soon.

claalmve: + test

src/containers/Tag/TagList/TagList.module.css
```diff
@@ -1,8 +1,87 @@
+.Buttons {
```
claalmve, line 1: I added a lot of styling, which wasn't necessarily needed for Pagination to work properly, but I'm always eager to use styling to make webpages look nice, so I figured I could add a little bit of it in my components here.

src/containers/Tag/TagList/TagList.tsx
```diff
@@ -44,66 +95,147 @@ export const TagList: React.SFC<TagListProps> = (props) => {
...
+                  <Divider orientation="vertical" />
+                  <CloseIcon className={styles.CloseIcon}></CloseIcon>
+                </div>
+              ) : null}
+            </div>
+          </form>
+
+          {/* Not sure how to override Material UI button styling without too much hassle */}
+          <Button
+            style={{
```
claalmve, line 199: As the comment says above, putting these styles the CSS file and importing them didn't work properly here (Material UI button overrode my styling). I'll be trying to figure out a solution to this, but in the meantime, the styling is exposed in this file.

src/containers/Tag/TagList/TagList.tsx
```diff
@@ -44,66 +95,147 @@ export const TagList: React.SFC<TagListProps> = (props) => {
...
+          data={tagList}
+          pagination
+          noHeader
+          onChangePage={(page, totalRows) => setPageNum(page)} // passes (page, totalRows)
+          onChangeRowsPerPage={(currentRowsPerPage, currentPage) => setPageRows(currentRowsPerPage)} // passes (currentRowsPerPage, currentPage)
+          paginationTotalRows={tagCount}
+          paginationServer
+          paginationPerPage={pageRows}
+          sortServer
+          // BUG: When trying to sort on a column, you have to click the column twice in order to see any changes.
```
claalmve, line 229: Very weird bug about sorting columns, may be related to the React-Data-Table package instead of my current functionality.

src/index.tsx
kurund: We need to have React.StrictMode. There is a lot of good reason why we need it. Check https://reactjs.org/docs/strict-mode.html. We should not be commenting it.

package.json
```diff
@@ -21,9 +22,11 @@
     "formik-material-ui": "^2.0.0-beta.1",
     "graphql": "^15.0.0",
     "react": "^16.13.1",
+    "react-data-table-component": "^6.9.3",
     "react-dom": "^16.13.1",
     "react-router-dom": "^5.2.0",
     "react-scripts": "3.4.1",
+    "styled-components": "^5.1.1",
```
kurund, line 29: Currently, our plan is to have CSS in separate files only. We should not add 'styled-components' package. The only exception is when we need to reuse some of the functions of them then we can add CSS in components.

src/containers/Tag/Tag.tsx
```diff
@@ -93,12 +93,12 @@ export const Tag: React.SFC<TagProps> = (props) => {
 
   const languageOptions = languages.data
     ? languages.data.languages.map((language: any) => {
-      return (
-        <MenuItem value={language.id} key={language.id}>
-          {language.label}
-        </MenuItem>
-      );
-    })
+        return (
```
kurund, line 96: Not sure why formatting is picked here. Can you please run "yarn format". Make sure to update your branch.

src/containers/Tag/TagList/TagList.module.css
kurund: Perfectly fine for now. The look and feel might change once we have design ready.

src/containers/Tag/TagList/TagList.tsx
kurund: Let's not add CSS to the component as explained above.

src/containers/Tag/TagList/TagList.tsx
kurund: OK great.

src/containers/Tag/TagList/TagList.tsx
kurund: Ok. I would suggest you post this issue in the "react-data-table" issue queue and see if they have any workaround for the same.

package.json
kurund: Looks like this is a dependency for react data table

src/components/pages/TagPage/TagPage.tsx
```diff
@@ -1,14 +1,11 @@
...
-import { Typography } from '@material-ui/core';
-
 import { TagList } from '../../../containers/Tag/TagList/TagList';
 
 export interface TagPageProps {}
 
 export const TagPage: React.SFC<TagPageProps> = () => {
   return (
     <div>
-      <Typography variant="h5">Tags</Typography>
```
claalmve, line 11: Did this to make title on same line as search and `Add New` (just for design purposes). I can change back if needed.

src/containers/Tag/TagList/TagList.tsx
```diff
@@ -53,67 +121,154 @@ export const TagList: React.SFC<TagListProps> = (props) => {
...
+            variant="contained"
+            disableElevation={true}
+            onClick={() => setNewTag(true)}
+          >
+            Add New
+          </Button>
+        </div>
       </div>
+
+      {/* Rendering list of tags */}
```
claalmve, line 250: This is table with client-side rendering (which will be removed later, just here for testing purposes).

src/containers/Tag/TagList/TagList.tsx
```diff
@@ -53,67 +121,154 @@ export const TagList: React.SFC<TagListProps> = (props) => {
...
-              <TableCell>Description</TableCell>
-              <TableCell></TableCell>
-            </TableRow>
-          </TableHead>
-          <TableBody>{listing}</TableBody>
-        </Table>
-      </TableContainer>
-    </div>
+
+      {/* Testing MaterialUI table */}
```
claalmve, line 260: Correct server-side pagination table using Material UI table components.

src/containers/Tag/TagList/TagList.tsx
```diff
@@ -53,67 +121,154 @@ export const TagList: React.SFC<TagListProps> = (props) => {
...
+
+  const handleSearch = (e: any) => {
+    e.preventDefault();
+    let searchVal = e.target.nameSearch.value.trim();
+    setSearchVal(searchVal);
+    resetTableVals();
+  };
+
+  // Making different columns for the table
+  const oldColumns = [
```
claalmve, line 179: Not needed with currently pagination implementation. Will be removed later (just used for testing purposes with client-side pagination component).

src/components/UI/ClientTable/ClientTable.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
```
claalmve, line 1: After looking into the different bugs with `DataTable` from `react-data-table-component`, I started from scratch and started using Material-UI table components instead, just to see if that would be a valid workaround. Now, in the `ServerTable.tsx` file, is where that component lies, with fully-functioning server-side pagination.

However, I found that `DataTable` was really easy to use and get up and running, so I decided to keep the component wrapping the DataTable around, as I think it is more advantageous than the Material UI table when it comes to client-side pagination or not having to worry about passing in so many props + maintaining their states. I named this one `ClientTable` and the other `ServerTable` for reference, and I think each one has its own advantages/disadvantages with respect to certain use cases.

Issue #38, opened 2020-06-15
Add unit tests for Form UI components


rathorevaibhav:
**Describe the task**
* add test cases for UI components targeted in #32 

**Expected behavior**
* All the test cases should pass

Issue #32, opened 2020-06-12
UI layer abstraction


rathorevaibhav:
**Describe the task**
When we switched to Formik, we need to create an abstraction layer to form UI components. This would basically be a wrapper of form fields targeting to put a better structure and cleaner components/containers.
```jsx
<div className={styles.Input}>
  <label className={styles.Label}>Label</label>
  <Field component={TextField} name="label" type="text" />
</div>
```

**Expected behavior**
1. The form fields to be moved to a structure like `UI > Layout > Form > Input`, `UI > Layout > Form > Checkbox` etc. These components will receive params like style, label, field type etc.
2. Update `Tag.tsx` component with the form and any other component to use these UI components.

PR #40, opened 2020-06-16
Button UI components

rathorevaibhav:
## Summary

Related to #32, added component for Button UI.

## Test Plan

To be covered in #38

### DIFF
--- a/src/containers/Tag/Tag.tsx
+++ b/src/containers/Tag/Tag.tsx
@@ -1,10 +1,10 @@
 import React, { useState, useEffect } from 'react';
 import { Redirect } from 'react-router-dom';
 import { Formik, Form, Field } from 'formik';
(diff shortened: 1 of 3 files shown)


src/containers/Tag/Tag.tsx
```diff
@@ -163,13 +163,9 @@ export const Tag: React.SFC<TagProps> = (props) => {
                 );
               })}
               <div className={styles.Buttons}>
-                <Button variant="contained" color="primary" onClick={submitForm}>
-                  Save
-                </Button>
+                <Button variant="contained" color="primary" onClick={submitForm} text="Save" />
```
kurund, line 166: I think instead of passing text we can send the children

rathorevaibhav: Done with the changes @kurund have a look

Issue #10, opened 2020-06-08
Implement a chat interface with WYSIWYG input


kurund:
**Describe the task**
We need to implement the chat interface based on the mockup in the requirement document.

**Expected behavior**
* Implement layout as per the mockup
* Ideally Whatsapp friendly editor
* We should support emoji's and bold to begin with.
* Evaluate: WYSIWYG https://www.npmjs.com/package/react-draft-wysiwyg for the input

**References**
* https://docs.google.com/document/d/1Q7ozYAHl1wcVQlXcg3iZjQZPHaUQJzQCYHTLaajncrs/edit?usp=sharing

kurund:
This issue is done and filing separate issue for formatting: https://github.com/glific/glific-frontend/issues/132

PR #60, opened 2020-06-21
Chat Interface work

kurund:
## Summary

* More work on Chat interface - #10 
* Added test cases for various components

## Test Plan

* Test cases are part of this PR


src/containers/Chat/ChatConversations/ChatConversation/ChatConversation.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+  test('it should render the name correctly', () => {
+    expect(wrapper.find('[data-testid="name"]').text()).toEqual('Jane Doe');
+  });
+
+  test('it should render the message content correctly', () => {
+    expect(wrapper.find('[data-testid="content"]').text()).toEqual('Hello there!');
+  });
+
+  test('it should render the message date correctly', () => {
+    expect(wrapper.find('[data-testid="date"]').text()).toEqual('00:14');
```
rathorevaibhav, line 30: is this the right time? considering it's 18:44 in the defaultProps

src/containers/Chat/ChatMessages/ChatInput/ChatInput.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    wrapper.find('[data-testid="send-button"]').simulate('click');
+    expect(inputSubmitted).toBeTruthy();
+  });
+
+  test('it should check if the enter is hit by user', () => {
+    const input = wrapper.find('[data-testid="message-input"]');
+    input.simulate('change', { target: { value: 'Hello' } });
+    input.simulate('keypress', { key: 'Enter' });
+
+    // TODO: both change and keypress are triggered correctly so wondering if we need any assertion here
```
rathorevaibhav, line 40: probably the same assertion as clicked on the button. The `inputSubmitted` should change to true. Thoughts?

src/containers/Chat/ChatConversations/ChatConversation/ChatConversation.test.tsx
kurund: yes, that's right! It is based on the timezone.

src/containers/Chat/ChatMessages/ChatInput/ChatInput.test.tsx
kurund: This is a key enter event and it not submitting the form directly. Hence "inputSubmitted" won't work.

src/containers/Chat/ChatConversations/ChatConversation/ChatConversation.test.tsx
rathorevaibhav: Hmm... I think we should avoid the timezone as now CI in place, it may have different timezone than IST and the test case may fail. What do you think?

kurund: I am not checking the actual data, I am checking mock data. So that should be fine. Check the props that are being sent.

PR #68, opened 2020-06-23
Adding Tests for ListIcon and Layout components/folders

claalmve:
## Summary
This PR is to provide additional testing for the ListIcon and Layout folders and the respective components within each directory.

## Test Plan
Making tests.


src/components/UI/Layout/Navigation/SideDrawer/SideDrawer.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+  );
+
+  it('initialized properly', () => {
+    const wrapper = mount(component);
+    expect(wrapper).toBeTruthy();
+    // Use Hidden at index 0 because there's two Drawer components.
+    expect(wrapper.find(Drawer).at(0).find(ListItem).length).toEqual(sideDrawerMenus.length);
+  });
+
+  it('callback is working', () => {
```
claalmve, line 31: This test isn't working yet. For some reason, `onClose` won't call and changed `isToggled` or print `being called`. Would love to get other insight about this as well.

kurund: We need to restructure SideDrawer according to new requirements. This won't be needed.

PR #69, opened 2020-06-23
TagList page test cases

rathorevaibhav:
## Summary

More test coverage for taglist page.

**Note covered**
- Test cases for filter tags and pagination


src/containers/Tag/Tag.test.tsx
```diff
@@ -1,11 +1,91 @@
...
+        },
+      },
+    },
+    result: {
+      data: {
+        createTag: {
+          tag: {
+            description: 'new Tag description',
+            id: '121',
+            isActive: true,
```
kurund, line 32: Isn't this should be false?

src/containers/Tag/Tag.test.tsx
```diff
@@ -1,11 +1,91 @@
...
+    request: {
+      query: FILTER_TAGS,
+      variables: {
+        filter: {
+          label: '',
+        },
+        opts: {
+          limit: 10,
+          offset: 0,
+          order: 'ASC',
```
kurund, line 68: Can you change this to DESC, so that it is consistent with the result.  I know it's mock data, but good to be correct.

PR #72, opened 2020-06-23
Error ToastMessage

cindyzhang12:
## Summary

- current not done the closing part of the message, useState does not work and am trying to do it through queries!
- change the ToastMessage component so that it takes in a vertical and horizontal prop which allows you to specify where you want it to show up on the screen
- changed all errors to ToastMessage popups with messages


kurund:
closing this in favor of #212

PR #74, opened 2020-06-24
Tooltip Component

cindyzhang12:
## Summary

- created a general tooltip component that can be used
- used tooltip component on ChatMessage.tsx so that the date will be shown on the right side of the message
- have not written tests for it yet, so it is not complete


src/components/UI/GeneralTooltip/GeneralTooltip.tsx
```diff
@@ -0,0 +1,16 @@
...
+interface GeneralTooltipProps {
+  title: String;
+  placement: TooltipProps['placement'];
+  body: any;
+}
+
+export const GeneralTooltip: React.SFC<GeneralTooltipProps> = (props: GeneralTooltipProps) => {
+  return (
+    <Tooltip title={props.title} placement={props.placement} arrow>
+      {props.body}
```
kurund, line 13: Let's use {props.children} here to make it cleaner.

src/components/UI/GeneralTooltip/GeneralTooltip.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import Tooltip, { TooltipProps } from '@material-ui/core/Tooltip';
+
+interface GeneralTooltipProps {
+  title: String;
+  placement: TooltipProps['placement'];
+  body: any;
```
kurund, line 7: We can remove body.

kurund:
@cindyzhang12 Can we just rename the component to just "Tooltip".

src/components/UI/GeneralTooltip/Tooltip.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import * as tooltip from '@material-ui/core/Tooltip';
```
kurund, line 2: Can we follow the convention used here: https://github.com/glific/glific-frontend/blob/master/src/components/UI/Form/Button/Button.tsx

kurund:
@cindyzhang12 Please rename the file to Tooltip.

src/components/UI/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  it('renders component properly', () => {
+    const wrapper = shallow(createTooltip());
+    expect(wrapper).toBeTruthy();
+  });
+
+  it('renders Tooltip', () => {
+    const wrapper = shallow(createTooltip());
+    expect(wrapper.find(Tooltip)).toBeTruthy();
+    console.log(wrapper.props());
```
rathorevaibhav, line 20: We should remove these debugging statements.

src/containers/Chat/ChatMessages/ChatMessage/ChatMessage.tsx
```diff
@@ -22,9 +23,11 @@ export const ChatMessage: React.SFC<ChatMessageProps> = (props) => {
 
   return (
     <div className={[styles.ChatMessage, additionalClass].join(' ')}>
-      <div className={styles.Content} data-testid="content">
-        {props.body}
-      </div>
+      <Tooltip title={moment(props.insertedAt).format('MMM, DD YYYY')} placement="right">
```
rathorevaibhav, line 26: is this working or should be this instead?
```suggestion
      <Tooltip title={moment(props.insertedAt).format('MM, DD YYYY')} placement="right">
```

rathorevaibhav: Also, need @kurund's input on this. Which date format are we using? I think it should be consistent system-wide.

cindyzhang12: The 'MMM' shows the first 3 letters of the month so it will look like Jan, Feb, Mar, etc. whereas 'MM' will show 1, 2, 3... 12. I can change this if we should be using a different format, please let me know what format you want! :)

PR #75, opened 2020-06-24
Feature/env

rathorevaibhav:
## Summary
Configured the env and config files

## Readme update
Updated Readme Installation steps for the project


kurund:
@rathorevaibhav Can you also move WebSocket URL to env

src/config.ts
```diff
@@ -0,0 +1,3 @@
+const envVariables = process.env;
```
kurund, line 1: May be move this file inside src/config

Issue #77, opened 2020-06-25
Changing Messages/Conversations GraphQL Endpoints


claalmve:
**Describe the task**
Change front-end GraphQL endpoints for chats because of changes in the endpoints from the back-end.

**Expected behavior**
Endpoints should work properly.

**References**
Specify any reference material / links

PR #78, opened 2020-06-25
Updating Conversation Queries

claalmve:
## Summary

Updating the conversation GraphQL endpoints. Addresses #77 .

## Test Plan

Passes all current tests in components using these queries.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -17,6 +17,7 @@ const App = () => {
           <Route path="/tag" exact component={TagPage} />
           <Route path="/tag/add" exact component={Tag} />
           <Route path="/tag/:id/edit" exact component={Tag} />
+          {/* Doesn't this error without a passed in `contactId`? */}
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -17,6 +17,7 @@ const App = () => {
           <Route path="/tag" exact component={TagPage} />
           <Route path="/tag/add" exact component={Tag} />
           <Route path="/tag/:id/edit" exact component={Tag} />
+          {/* Doesn't this error without a passed in `contactId`? */}
```
claalmve, line 20: Because I was getting an error page on the default `/chat` route, I did a little digging into it, and I believe because contactId is never passed in for this route, the page will not render properly. So, I'm not completely sure how this worked before, or if I did something wrong on my end, but now to see the page properly, you need to add a ID at the end of the URL (i.e. localhost:3000/chat/2)

PR #80, opened 2020-06-25
UI cleanup

kurund:
## Summary

UI Clean up and move towards final design 

## Test Plan

Fixes included in this PR


src/containers/Chat/ChatConversations/ChatConversation/ChatConversation.tsx
```diff
@@ -40,7 +35,7 @@ const ChatConversation: React.SFC<ChatConversationProps> = (props) => {
           {props.lastMessage.body}
         </div>
         <div className={styles.MessageDate} data-testid="date">
-          {moment(props.lastMessage.insertedAt).format('HH:mm')}
+          {moment(props.lastMessage.insertedAt).format('D/M/Y')}
```
rathorevaibhav, line 38: I think we need a middle ground here to show the message time and somehow tell users what day/date they came in.
That can be a separate task I believe.

kurund: I am bit on fence for this. I think showing date is fine here. They can see the time when they click on the conversation. We can check Abhishek at some point on his view. For now we can keep this.

rathorevaibhav: Yep, I agree. This is fine for now.

Issue #85, opened 2020-06-25
Set selected / highlight current route in the sidebar


kurund:
**Describe the task**
Implement highlighting for the sidebar navigation menu.

**Expected behavior**
If we are navigating to /chat then the "Chats" in the sidebar should be highlighted

PR #89, opened 2020-06-25
Show MenuItem Selected

claalmve:
## Summary

Shows which MenuItem is selected. Addresses #85.


src/components/UI/Layout/Navigation/SideMenus/SideMenus.tsx
```diff
@@ -8,9 +8,17 @@ import ListIcon from '../../../ListIcon/ListIcon';
 export interface SideMenusProps {}
 
 const SideMenus: React.SFC<SideMenusProps> = () => {
-  const menuList = sideDrawerMenus.map((menu) => {
+  const [selectedIndex, setSelectedIndex] = React.useState(0);
```
kurund, line 11: Let's same "useState" convention that we use in other places to be consistent.

PR #113, opened 2020-07-01
Fixing no contacts bug

claalmve:
## Summary

Small bug that occurs when there are no contacts initially, and so, no conversations.


src/containers/Chat/ChatMessages/ChatMessages.tsx
```diff
@@ -328,10 +330,17 @@ export const ChatMessages: React.SFC<ChatMessagesProps> = ({ contactId }) => {
     <Container className={styles.ChatMessages} disableGutters>
       {dialogBox}
       {toastMessage}
-      <ContactBar contactName={conversations.contact.name} />
-      <Container className={styles.MessageList} data-testid="messageContainer">
-        {messageList}
-      </Container>
+      <ContactBar contactName={conversations ? conversations.contact.name : 'No contacts'} />
+      {/* To avoid warnings with empty children in Container */}
```
claalmve, line 334: I decided switching between element types would allow for greatest flexibility, as `Container` throws a warning when children are `undefined`.

src/containers/Chat/ChatMessages/ChatMessages.tsx
```diff
@@ -328,10 +330,17 @@ export const ChatMessages: React.SFC<ChatMessagesProps> = ({ contactId }) => {
...
+      {messageList ? (
+        <Container className={styles.MessageList} data-testid="messageContainer">
+          {messageList}
+        </Container>
+      ) : (
+        <div className={styles.MessageList} data-testid="messageContainer">
+          {messageList}
```
rathorevaibhav, line 341: I think the Container & div pairing is fine. Just one question, when there's nothing in the message list, shouldn't we be showing an appropriate message to the user. Like `No messages for this conversation` or something?

src/containers/Chat/ChatMessages/ChatMessages.tsx
```diff
@@ -287,7 +287,9 @@ export const ChatMessages: React.SFC<ChatMessagesProps> = ({ contactId }) => {
   }
 
   let messageList: any;
-  if (conversations.messages.length > 0) {
+
+  // Must have a contact + conversations in order for messages to render.
```
rathorevaibhav, line 291: didn't get this comment completely. When we say "contact", what are we referring to exactly inside the if statement?

src/containers/Chat/ChatMessages/ChatMessages.tsx
claalmve: Added this message.

claalmve:
Haven't been able to totally check this with conversations as well because my back-end is falling apart. Will check on it once I get this fixed.

claalmve:
^ Resolved.

PR #114, opened 2020-07-01
Registration and Authentication

cindyzhang12:
## Summary

- create registration page
- integrated authentication
- still have to do:
   - errors on form submission
   - submit for auth page

PR #152, opened 2020-07-07
Login Branch from Registration Branch

cindyzhang12:
## Summary

- this branch branches off the registration branch since it is necessary to create users for login


src/components/pages/ConfirmOTP/ConfirmOTP.tsx
```diff
@@ -0,0 +1,57 @@
...
+    props.location ? props.location.state.authCode : undefined
+  );
+
+  const handleuserAuthCodeChange = () => (event: React.ChangeEvent<HTMLInputElement>) => {
+    setuserAuthCode(event.target.value);
+  };
+
+  const handleSubmit = () => {
+    if (userAuthCode == correctAuthCode) {
+      window.location.href = REACT_APP_GLIFIC_CHAT_REDIRECT;
```
cindyzhang12, line 27: Is this method of redirecting the user to the chat page okay? Is there a better way to do it?

kurund: You can use <Redirect> to send to chat page.

src/config/axios.ts
```diff
@@ -0,0 +1,7 @@
+export const REACT_APP_GLIFIC_REGISTRATION_API =
+  process.env.REACT_APP_GLIFIC_API + '/v1/registration';
+export const REACT_APP_GLIFIC_AUTHENTICATION_API =
+  process.env.REACT_APP_GLIFIC_API + '/v1/registration/send_otp';
+export const REACT_APP_GLIFIC_CHAT_REDIRECT = 'http://localhost:3000/chat';
```
kurund, line 5: We don't need this.

src/common/constants.ts
```diff
@@ -1,3 +1,10 @@
 export const SIDE_DRAWER_WIDTH = 200;
 export const DATE_FORMAT = 'DD/MM/YY';
 export const TIME_FORMAT = 'HH:mm';
+export const REACT_APP_GLIFIC_REGISTRATION_API =
+  process.env.REACT_APP_GLIFIC_API + '/v1/registration';
+export const REACT_APP_GLIFIC_AUTHENTICATION_API =
+  process.env.REACT_APP_GLIFIC_API + '/v1/registration/send_otp';
+export const REACT_APP_GLIFIC_CHAT_REDIRECT = 'http://localhost:3000/chat';
```
kurund, line 8: We should not hardcode this here. I don't think we need redirect URL. We can directly send the user to the route.

kurund:
@cindyzhang12 Can you please check the comments on https://github.com/glific/glific-frontend/pull/114 and just add it in this branch.

Issue #132, opened 2020-07-05
Formatting support for chat message input


kurund:
**Describe the task**
* Support basic text message formatting such as bold.

**Expected behavior**
* Message sent should be displayed as per formatting on Whatsapp and chat interface

claalmve:
Check out how WhatsApp deals with these inputs (i.e. *bold* for bold)

claalmve:
- https://draftjs.org/docs/api-reference-content-state
- https://www.npmjs.com/package/markdown-draft-js

claalmve:
- https://www.npmjs.com/package/marked for rendering Markdown back to HTML

Issue #126, opened 2020-07-03
Message input to support multiple lines


rathorevaibhav:
**Describe the task**
A user can send multiple lines within a message. Right now, we cannot send more than one line in a message.

**Expected behavior**
1. When typing `Shift + Enter` from the keyboard, the message input should move the cursor to a new line without sending the message.
2. When typing only `Enter` from the keyboard, the message should be sent like it's happening now.

**References**
See WhatsApp Web to refer the functionality

claalmve:
Use textarea.

PR #166, opened 2020-07-08
Adding additional functionality for messaging + rendering messages

claalmve:
## Summary

Allows for more rich functionality for messaging (bold, italics, shift-enter), and renders these styled texts in the messages window as well. Addresses #126 and #132.

Nice to adds:
- Ability for chatbox to expand upwards on overflow (like what WhatsApp does). Not totally sure how to get the two components to resize dynamically on text input size.
- Fixing small bug: When bolding or applying styling over highlighted text, the cursor must move at least once in order for it to send the message correctly.


src/containers/Chat/ChatMessages/ChatInput/ChatInput.tsx
```diff
@@ -52,46 +56,76 @@ export const ChatInput: React.SFC<ChatInputProps> = ({ onSendMessage }) => {
     );
   }
 
+  const handleChange = (editorState: any) => {
+    // Get the markdown equivalent for this text.
+    // BUG: When highlighting over text and bolding, the cursor must move at least once in order for the changes to take effect.
+    let markdownString = draftToMarkdown(convertToRaw(editorState.getCurrentContent()));
```
claalmve, line 62: The `getCurrentContent` method causes two tests to fail. I tried looking up the error but I haven't had much success with it. Will continue trying to debug later, but would love to get help on passing this.

Current error: `TypeError: editorState.getCurrentContent is not a function`

rathorevaibhav:
@claalmve I did some changes to `ChatInput.test.tsx` to pass the test cases. Can you take a look at the code and see if it's making sense?

PR #172, opened 2020-07-09
Display phone number in conversation list and on chat if name is not received

cindyzhang12:
## Summary

- shows phone number for conversation if contact name doesn't exist


kurund:
@cindyzhang12 Please add the test case for this.

kurund:
Code coverage and tests are handled in a separate issue.

Issue #154, opened 2020-07-07
Search ChatConversations bugs


claalmve:
**Describe the bugs**

- [x] 1. Resetting search sends a search request with an empty string that returns all conversations objects (making it slow to render back on an initial reset).
- [x] 2. Resetting search also doesn't render correct messages in the window corresponding to the conversations in the list (i.e. Default receiver messages being shown but Rossie Bruen's convo indicated as currently clicked on in convo list). Has to do with `selectedIndex` in ChatConversations state.
- [x] 3. When searching for individual that doesn't currently exist in `allConversations` list, additional query is made in ChatMessages file and causes re-render of ChatConversations list, and incorrect conversations to be listed. May be solved with additional caching for filtered conversations?
- [x] 4. When searching for some individuals that already exist in the conversations cache and clicking on their respective conversation, the messages are correctly rendered but the search/filtering of conversations is reset, which in turn resets the conversation list to be incorrect.

**The last 3 issues seem to do with a re-render of the page causing the current `searchVal` to be reset.**

claalmve:
#138 is addressed here as the 3rd and 4th issues (for both cases where the conversation is in the cache and is not).

claalmve:
1. Semi-solved, with search of null returning limited responses. **However**, search should not even be called on a reset search.

claalmve:
**When searching for Boris on the INITIAL chat page (no contact id in URL) and clicking on Boris, the page reloads.** Similar behavior follows on the **INITIAL** chat page.

Issue #157, opened 2020-07-07
Chat searching, the results should filter real time


kurund:
**Is your feature request related to a problem? Please describe.**
* Usability improvement

**Describe the solution you'd like**
* For chat searching, the results should filter real time. this would remove the requirement of 'Enter' or Submit action

claalmve:
On keyup, wait, and then hit the endpoint. Set timeout function. When the user stops for brief amount of time (0.5 seconds?), make this request.

kurund:
Sounds good to me.

claalmve:
https://github.com/slorber/react-async-hook#how-can-i-implement-a-debounced-search-input--autocomplete

PR #182, opened 2020-07-10
Search Bugs

claalmve:
## Summary

Solving search bugs with ChatConversations in #154 .

## Test Plan

WIP.


src/containers/Chat/ChatConversations/ChatConversations.tsx
```diff
@@ -114,6 +114,19 @@ export const ChatConversations: React.SFC<ChatConversationsProps> = () => {
     setSelectedIndex(-1);
   };
 
+  const handleChange = (e: any) => {
```
claalmve, line 117: Related to #157 . Put them both in the same branch for now since they're both related to search. Can split it later if I can't find a quick fix for real-time search/the fix is large.

src/containers/Chat/ChatConversations/SearchBar.tsx
```diff
@@ -31,7 +32,12 @@ export const SearchBar: React.SFC<SearchBarProps> = (props) => {
               className={styles.SearchField}
               name="searchInput" // This is important for extracting the search value in parent component.
               placeholder="Search"
-              onChange={(e) => setLocalSearchVal(e.target.value)}
+              onChange={(e: any) => {
```
claalmve, line 35: Same as above.

src/containers/Chat/ChatMessages/ChatMessages.tsx
```diff
@@ -224,17 +225,19 @@ export const ChatMessages: React.SFC<ChatMessagesProps> = ({ contactId }) => {
...
-  if (conversationInfo.length === 0) {
-    conversationIndex = 0;
-    // No conversations case
     if (allConversations.conversations.length === 0) {
       conversationInfo = null;
     } else {
-      conversationInfo = allConversations.conversations[conversationIndex];
+      // POTENTIAL FIX: If no contactId, redirect to URL with contactId.
+      // This could be on a higher component (App/Chat) instead of here. However, this is where we first know the contactId.
+      return <Redirect to={'/chat/'.concat(allConversations.conversations[0].contact.id)} />;
```
claalmve, line 237: This is currently my potential fix for this issue. After seeing that all these bugs are primarily focused on changing the chat URL from an empty contactId to an existing one, I thought about always redirecting to a conversation with a contactId instead. I also took a look at how other chat applications have done with related functionality (Messenger), and I realized that their URLs always redirect to the most recent conversation + contactId (similar to what I've implemented here).

I don't know if this is the best solution for this issue, but this + the additional fix on line 232 would ensure that there would never be a need for a user to enter the chat endpoint without a contactId and come across an error.

src/containers/Chat/Chat.tsx
```diff
@@ -121,6 +121,9 @@ const Chat: React.SFC<ChatProps> = ({ contactId }) => {
     return null;
   }
 
+  // Other location.
```
claalmve, line 124: This is referenced from the last comment. The functionality from there could be moved one file up (in Chat.tsx instead of ChatMessages.tsx), but just wanted to get some feedback into whether or not this would be a better place for it.

claalmve:
For this PR, I decided to get around these bugs by redirecting to a URL with an existing contactId. I've written more about this fix in a comment on the line fix (line 237), but I think that though this doesn't directly address the issue with re-rednering search, I do believe this would also be a valid solution, since #1 this bug would always be avoided and #2 this is how similar chat platforms perform as well.

I'd love to get feedback on this decision as well, as I figured this might be the quickest solution but not necessarily the most comprehensive.

kurund:
@claalmve Yes, we can always append the contact id in the URL and let's move it to one level up. Chats.ts

src/containers/Chat/Chat.tsx
kurund: Yes, let's do it.

src/containers/Chat/ChatConversations/ChatConversations.tsx
kurund: Yes, that's fine. Both can be in this branch.

src/containers/Chat/ChatMessages/ChatMessages.tsx
kurund: Yes, that's fine. Let's do it.

src/containers/Chat/ChatConversations/ChatConversations.tsx
kurund: As discussed this is now part of separate PR https://github.com/glific/glific-frontend/pull/191

src/containers/Chat/Chat.tsx
```diff
@@ -115,11 +119,32 @@ const Chat: React.SFC<ChatProps> = ({ contactId }) => {
   if (loading) return <Loading />;
   if (error) return <p>Error :(</p>;
 
+  const getConversations = () => {
+    const allConversations: any = client.readQuery({
```
kurund, line 123: Why are we doing this? The query is fired in this component itself to get the conversations. We don't need to read it from the cache.

src/containers/Chat/Chat.tsx
```diff
@@ -115,11 +119,32 @@ const Chat: React.SFC<ChatProps> = ({ contactId }) => {
...
+  }
+
   return (
     <Paper>
       <div className={styles.Chat}>
         <div className={styles.ChatMessages}>
-          <ChatMessages contactId={contactId} />
+          <ChatMessages
+            contactId={newContactId}
+            getConversations={getConversations}
```
kurund, line 145: We should not be doing this here. See above comment.

src/containers/Chat/ChatConversations/SearchBar.tsx
```diff
@@ -31,7 +32,12 @@ export const SearchBar: React.SFC<SearchBarProps> = (props) => {
...
+                setLocalSearchVal(e.target.value);
+                // if (props.handleChange) {
```
kurund, line 37: Please remove unwanted code.

src/containers/Chat/ChatMessages/ChatMessages.tsx
```diff
@@ -409,7 +402,7 @@ export const ChatMessages: React.SFC<ChatMessagesProps> = ({ contactId }) => {
   // Check if there are conversation messages else display no messages
   if (messageList) {
     messageListContainer = (
-      <Container className={styles.MessageList} data-testid="messageContainer">
+      <Container className={styles.MessageList} maxWidth={false} data-testid="messageContainer">
```
kurund, line 405: Why do we need this?

src/containers/Chat/ChatMessages/ChatMessages.tsx
```diff
@@ -64,13 +61,23 @@ interface ChatMessagesInterface {
   };
 }
 
+export interface ChatMessagesProps {
+  contactId: number;
+  getConversations: () => any;
```
kurund, line 66: This might change based on my comment on Chats.ts

PR #191, opened 2020-07-14
Dynamic Searching for Conversations

claalmve:
## Summary

Implementing dynamic searching so that results will render while the user is typing.

## Test Plan

WIP.


kurund:
@claalmve Looks like this PR has gone stale and some of the code was part of another PR. Can you please create a clean PR for the changes specific to dynamic search and close this.

PR #192, opened 2020-07-14
Feature/authentication

rathorevaibhav:
## Summary

1. Authentication module.
2. Public and private routes

## Test Plan

Test cases added

## Note
One test case in App.test.tsx is failing in GitHub CI due to an error related to Websocket. I've commented it until I find a suitable fix for that.
image


src/common/session.ts
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
```
kurund, line 1: Context is kind of special and different from other files in this folder.  So should we create "src/context" folder and move this there?

PR #193, opened 2020-07-14
Error for if phone number already exists when registering user

cindyzhang12:
## Summary
- shows error on confirm otp page if phone number already exists

image


kurund:
@cindyzhang12 Can we please add a test for this too. Thank you.

PR #198, opened 2020-07-15
Code cleanups and refactoring

rathorevaibhav:
## Summary

Code cleanups based on CodeBeat feedbacks

### DIFF
--- a/src/containers/List/List.tsx
+++ b/src/containers/List/List.tsx
@@ -206,9 +206,8 @@ export const List: React.SFC<ListProps> = ({
     }
   }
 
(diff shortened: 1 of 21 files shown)


src/containers/List/List.tsx
```diff
@@ -0,0 +1,291 @@
...
+            onClick={() => showDialogHandler(id!, label)}
+          >
+            <DeleteIcon />
+          </IconButton>
+        </div>
+      );
+    }
+  }
+
+  function formatList(tags: Array<any>) {
```
kurund, line 209: Tags reference should be removed from this.

PR #213, opened 2020-07-20
HSM Page

cindyzhang12:
## Summary
- need to add icons


src/components/pages/HSMPage/HSMPage.tsx
```diff
@@ -0,0 +1,49 @@
...
+    columnNames: columnNames,
+    columns: getColumns,
+    columnStyles: columnStyles,
+  };
+
+  return (
+    <List
+      title="HSM Messages"
+      listItem="sessionTemplates"
+      listItemName="speed send"
```
kurund, line 41: This seems incorrect too. This should be "template". Please update the code when you copy+paste

src/App.tsx
```diff
@@ -43,6 +44,9 @@ const App = () => {
             <Route path="/speed-send/add" exact component={MessageTemplate} />
             <Route path="/speed-send/:id/edit" exact component={MessageTemplate} />
             <Route path="/chat" exact component={Chat} />
+            <Route path="/templates" exact component={HSMPage} />
```
kurund, line 47: Please use singular so "template"

src/App.tsx
```diff
@@ -43,6 +44,9 @@ const App = () => {
...
+            {/* This part isn't working properly */}
```
kurund, line 48: Why is this added? Please clean up unnecessary code.

cindyzhang12: That comment was there before, I wasn't sure who added it

PR #227, opened 2020-07-25
Saved search makes an entry

kurund:
## Summary

* Add Saved Search Toolbar

## Test Plan

* Included in the PR


kurund:
@rathorevaibhav merging this as my next work is dependent on this.

PR #280, opened 2020-08-01
Feature/table footer fixed

rathorevaibhav:
## Summary

Position fixed for pagination tab and transparent background. make it stick to the bottom in case there are rows more than visible window height.

## Test Plan

Not needed


src/components/UI/Pager/Pager.tsx
```diff
@@ -141,12 +142,19 @@ export const Pager: React.SFC<PagerProps> = (props) => {
     props.tableVals
   );
 
+  useEffect(() => {
+    var html = document.querySelector('html');
+    if (html && html.scrollHeight > html.clientHeight) {
+      setTableFooterStyle(undefined);
```
kurund, line 148: Setting "undefined" does not look correct. Can you please explain the logic.

PR #293, opened 2020-08-04
Code coverage

kurund:
## Summary

* Increase test coverage

## Test Plan

* Included

### DIFF
--- a/src/containers/Template/Form/HSM/HSM.test.tsx
+++ b/src/containers/Template/Form/HSM/HSM.test.tsx
@@ -37,7 +37,7 @@ test('check for validations for the HSM form', async () => {
 
   fireEvent.change(container.querySelector('input[name="label"]'), {
     target: {
-      value: 'We are not allowing an really long title, and we should trigger validtaion for this.',
+      value: 'We are not allowing a really long title, and we should trigger validation for this.',
     },
   });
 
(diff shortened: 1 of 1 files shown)


src/containers/Template/Form/HSM/HSM.test.tsx
```diff
@@ -1,26 +1,46 @@
...
+  const button = queryByText('Save');
+  fireEvent.click(button);
+  await wait();
+
+  // we should have 2 errors
+  expect(getAllByText('Required')).toHaveLength(2);
+
+  fireEvent.change(container.querySelector('input[name="label"]'), {
+    target: {
+      value: 'We are not allowing an really long title, and we should trigger validtaion for this.',
```
rathorevaibhav, line 40: some type here
```suggestion
      value: 'We are not allowing a really long title, and we should trigger validation for this.',
```

src/containers/Template/Form/HSM/HSM.test.tsx
```diff
@@ -1,26 +1,46 @@
...
+    },
+  });
+
+  // we should still have 2 errors
+  expect(getAllByText('Required')).toHaveLength(2);
```
rathorevaibhav, line 45: will this test case change once we fix the error message that need to be displayed?

kurund: Yes, this is a temporary fix as per current behavior.

PR #319, opened 2020-08-06
Links in messages

rathorevaibhav:
## Summary

1. Link support in messages. outgoing and incoming.
2. Some cleanups in the way message string conversion was done for bold, italics etc.

## Test Plan

Added


src/react-app-env.d.ts
```diff
@@ -1 +1,3 @@
 /// <reference types="react-scripts" />
+
+declare module 'react-tiny-link'
```
kurund, line 3: We should move this to types folder

src/containers/Chat/ChatMessages/ChatMessage/ChatMessage.tsx
```diff
@@ -162,14 +164,46 @@ export const ChatMessage: React.SFC<ChatMessageProps> = (props) => {
     />
   );
 
+  let linkPreview = null;
```
kurund, line 167: If possible can we abstract this code and move out of ChatMessage.

PR #323, opened 2020-08-11
Yup fixes

kurund:
## Summary

* Let's use Yup package for common form validations

## Test Plan

* Included in the PR


src/containers/List/ListItem/ListItem.tsx
```diff
@@ -42,7 +42,8 @@ export const ListItem: React.SFC<ListItemProps> = ({
   deleteItemQuery,
   states,
   setStates,
-  setValidation,
+  //setValidation,
```
rathorevaibhav, line 45: should we remove if not needed?

PR #331, opened 2020-08-14
Create collection

DigneshGujarathi:


src/containers/Form/FormLayout.tsx
```diff
@@ -110,13 +115,45 @@ export const FormLayout: React.SFC<FormLayoutProps> = ({
     return null;
   }
 
+  const collectionPayload = (payload: any) => {
```
kurund, line 118: We should try to move this to the collection component.

src/containers/Form/FormLayout.tsx
```diff
@@ -98,6 +100,9 @@ export const FormLayout: React.SFC<FormLayoutProps> = ({
       if (!itemId) setLink(data[`create${camelCaseItem}`][listItem][linkParameter]);
       setFormSubmitted(true);
     },
+    onError: (error: ApolloError) => {
+      console.log('Error', error);
```
kurund, line 104: We should use error component here.

PR #333, opened 2020-08-17
Added edit collection & code cleanup

DigneshGujarathi:

### DIFF
--- a/src/components/UI/Form/Calendar/Calendar.tsx
+++ b/src/components/UI/Form/Calendar/Calendar.tsx
@@ -6,6 +6,7 @@ import DateFnsUtils from '@date-io/date-fns';
 import { MuiPickersUtilsProvider, KeyboardDatePicker } from '@material-ui/pickers';
 import { getIn } from 'formik';
 import moment from 'moment';
+import { DATE_FORMAT } from '../../../../common/constants';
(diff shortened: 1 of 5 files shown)


src/components/UI/Form/Calendar/Calendar.tsx
```diff
@@ -27,8 +27,8 @@ export const Calendar: React.SFC<CalendarProps> = ({
   const errorText = getIn(errors, field.name);
   const touchedVal = getIn(touched, field.name);
   const hasError = dirty && touchedVal && errorText !== undefined;
-
-  const dateValue = field.value ? field.value : new Date();
+  moment.defaultFormat = 'DD.MM.YYYY HH:mm';
```
kurund, line 30: Please use const defined instead of "DD.MM.YYYY HH:mm"

src/containers/Form/FormLayout.tsx
```diff
@@ -153,7 +127,7 @@ export const FormLayout: React.SFC<FormLayoutProps> = ({
 
     // create custom payload for collection
     if (listItemName === 'collection') {
```
kurund, line 129: Maybe change this to if "setPayload", then use that.

PR #334, opened 2020-08-17
Feature/add contact to group

rathorevaibhav:
## Summary

Add groups for a contact

## Test Plan

Added


kurund (commented):
@rathorevaibhav We need to change the UI to use autocomplete interface here too.

PR #368, opened 2020-08-28
Added update & create collection button

DigneshGujarathi:


kurund:
@DigneshGujarathi 
Found a few issues:
- [ ] setting default not working when search and modify search criteria
- [ ] search does not refresh result set

kurund:
@DigneshGujarathi Found an issue when we save a collection and then save another one using create new.
Screenshot 2020-09-01 at 8 31 40 PM

src/containers/Collection/Collection.tsx
```diff
@@ -173,11 +250,12 @@ export const Collection: React.SFC<CollectionProps> = ({ match, type, search, ..
   const setPayload = (payload: any) => {
     if (search) search(payload);
     if (props.searchParam) {
-      payload.term = props.searchParam.term;
-      payload.includeTags = props.searchParam.includeTags;
-      payload.includeGroups = props.searchParam.includeGroups;
-      payload.dateTo = props.searchParam.dateTo;
-      payload.dateFrom = props.searchParam.dateFrom;
+      // payload.term = props.searchParam.term;
```
kurund, line 253: We should remove this code if not needed.

PR #369, opened 2020-08-28
Feature/draft js in templates

rathorevaibhav:
## Summary

DraftJS support in speed send

## Test Plan

Added


kurund (changes requested):
Noticed a few issues that needs be fixed.
- [x] In Edit mode message is displayed in Markdown mode, it should be same as how we add them.
- [x] Noticed slight slowness when typing messages in the Messages field
- [ ] Height of chat input has increased on Chat screen. ( Not sure if it was intentional )

kurund (approved):
Looks good. There are some improvements needed, I will file a new issue.

PR #373, opened 2020-08-29
Update profile for a contact

rathorevaibhav:
## Summary

Users can now update contact details from the chat screen. Two fields are purposely left out:

- Contact settings
- Contact fields

These fields will be added when there will be a better idea from the design inputs and the details that'll get saved in these.

## Test Plan

Added


kurund (approved):
@rathorevaibhav Looks good. We need to handle delete action better and redirect to "/chat". Will file separate issue

kurund:
Here you go https://github.com/glific/glific-frontend/issues/376

PR #374, opened 2020-08-29
Showing session timer in chat window

rathorevaibhav:
## Summary

Showing session timer in the chat window

## Test Plan

Added


kurund (approved):
@rathorevaibhav We should be using "lastMessageAt" from contact instead of manually calculating the same. I have updated the code for the same.

Issue #366, opened 2020-08-28
Page refresh user is redirected to chat screen


kurund:
**Describe the bug**
* Visit any page other than chat and then refresh. User is redirected to chat page.

**To Reproduce**
1. Go to 'http://localhost:3000/tag'
2. Refresh the page
3. User is redirected to  'http://localhost:3000/chat'

**Expected behavior**
* User should remain on http://localhost:3000/tag

PR #377, opened 2020-08-30
Auth cleanup

kurund:
## Summary

- [x] Fixes the issue mentioned #366 
- [x] If invalid token then we auto-renew using the refresh token
- [x] if invalid token and refresh token just logout the user

## Test Plan

* Not needed


rathorevaibhav (commented):
reviewed and tested. looking good.

PR #387, opened 2020-08-31
Add automations to group and user

rathorevaibhav:
## Summary

1. feature to add automation to a group
2. feature to add automation for a user
2. needed same component in both so had to work in a single branch

## Test Plan

Added


kurund (approved):
Looks good.

PR #397, opened 2020-09-01
Bug fixed for group not showing

rathorevaibhav:
## Summary

Groups should show properly now when assigning to a contact.


kurund (approved):
works well!

PR #400, opened 2020-09-01
Hierarchy of tags

DigneshGujarathi:


src/containers/Tag/Tag.tsx
```diff
@@ -34,19 +37,47 @@ export const Tag: React.SFC<TagProps> = ({ match }) => {
...
   };
 
+  const { data } = useQuery(GET_TAGS);
+
+  if (!data) return <Loading />;
+
+  let tags = [];
+  // remove the self tag from list
+  if (data && match && match.params.id) {
+    tags = data.tags.filter(function (tag: any) {
```
kurund, line 60:
```suggestion
    tags = data.tags.filter((tag: any) =>  {
```

We should use ES6 syntax

kurund (approved):
added minor correction

kurund (approved):
Looks good.

PR #405, opened 2020-09-02
Contact block functionality

rathorevaibhav:
## Summary

Block a contact functionality.

## Test Plan

Added


kurund (changes requested):
@rathorevaibhav Looks good, one small change is needed. After blocking we should redirect to "/chat"

kurund (approved):

PR #422, opened 2020-09-03
Session timer bug for new messages

rathorevaibhav:
## Summary

1. Timer will update as soon as we get a new message.
2. Also contains fixes for T0 when session times out.

## Test Plan

TBD


kurund (approved):
Looks good!

PR #435, opened 2020-09-06
Code improvements

kurund:
## Summary

* Improve code coverage
* Remove duplication

## Test Plan

* Update existing and added new test cases

### DIFF
--- a/src/services/AuthService.tsx
+++ b/src/services/AuthService.tsx
@@ -64,6 +64,7 @@ export const checkAuthStatusService = () => {
   return authStatus;
 };
 
+// service to sent the OTP based on the phone number
 export const sendOTP = (phoneNumber: string) => {
   return axios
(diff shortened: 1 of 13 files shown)


src/services/AuthService.tsx
```diff
@@ -64,18 +64,19 @@ export const checkAuthStatusService = () => {
   return authStatus;
 };
 
-export const sendOTP = (phoneNumber: string) => {
+// service to sent the OTP based on the phone number
+export const sendOTP = (phoneNumber: string, registration = 'false') => {
```
rathorevaibhav, line 68: any reason for going with string instead of boolean for the registration param?

kurund: The backend needs it that way,

rathorevaibhav (approved):

PR #444, opened 2020-09-08
Editor enhancements according to feedbacks

rathorevaibhav:
## Summary

- [x]  Remove underline, as it is not supported by Whatsapp.
- [x]  We should support: https://faq.whatsapp.com/general/chats/how-to-format-your-messages/?lang=fb
- [x]  Spacing between 2 paragraphs is currently removed
- [x]  On the chat screen, when we select speed send or template, it is populated incorrectly. spacing etc is removed.
- [x]  On add/edit speed send clicking on the emoticon icon shows the option to select emoticons. It should close pop up if you click anywhere outside.


kurund (approved):
Looks good!

PR #447, opened 2020-09-09
Added user role & permissions for layout

DigneshGujarathi:


src/App.tsx
```diff
@@ -52,60 +54,89 @@ const App = () => {
   };
 
   let routes;
+  if (userRole.length == 0) {
+    // on page refresh wait to get user role
+    setTimeout(() => {
```
kurund, line 59: I am not sure if setTimeout is a good option here.

src/App.tsx
```diff
@@ -52,60 +54,89 @@ const App = () => {
...
-            <Route path="/speed-send/:id/edit" exact component={SpeedSend} />
-            <Route path="/automation" exact component={AutomationList} />
-            <Route path="/automation/add" exact component={Automation} />
-            <Route path="/automation/:id/edit" exact component={Automation} />
-            <Route path="/group" exact component={GroupList} />
-            <Route path="/group/add" exact component={Group} />
-            <Route path="/group/:id/edit" exact component={Group} />
-            <Route path="/group/:id/contacts" exact component={GroupContact} />
+    let route;
+    if (userRole.includes('Staff')) {
```
kurund, line 67: I think we should move routes out of app.tsx, it's getting cluttered. We could potentially just get routes by role method included here.

src/components/UI/Layout/Navigation/SideDrawer/SideDrawer.tsx
```diff
@@ -109,6 +110,15 @@ export const SideDrawer: React.SFC<SideDrawerProps> = (props) => {
   const [mobileOpen, setMobileOpen] = useState(false);
   const [fullOpen, setFullOpen] = React.useState(true);
   const [active, setActive] = React.useState(false);
+  const [staffManagementMenus, setStaffManagementMenus] = React.useState<any>([]);
+
+  // on page reload check menus for user role
+  if (staffManagementMenus.length === 0) {
+    // Need to wait before you get the role from API
+    setTimeout(() => {
```
kurund, line 118: Consider an alternative for setTimeout

kurund (changes requested):

kurund (approved):
@DigneshGujarathi Finally finished reviewing the PR. Overall looks good. I am merging this PR.

2 Issues that could be addressed in separate PR.
* Manager should not be able to update their role to Admin
* For staff role, we can see the groups but not the contacts in that group. We should be able to view the contacts too.

PR #448, opened 2020-09-09
Feature/add media support

rathorevaibhav:
## Summary

Support for the following media files:
- [x] Audio
- [x] Video
- [x] Image
- [x] Document

## Test Plan

Added


kurund (changes requested):
Can we please rename "videoThumbnail.jpeg" to "videothumbnail.jpeg" so we are consistent.

kurund:
@rathorevaibhav test with audio, video, images, pdf Looks good. Just minor comments.

PR #451, opened 2020-09-10
Remove shortcode

rathorevaibhav:
## Summary

Shortcode support in automation is removed from the backend.
So removing its support from frontend.

## Test Plan

Updated

Need this for PR #448


kurund (approved):
Looks good.

PR #455, opened 2020-09-10
Added multi select functionality in search tag

rathorevaibhav:
## Summary

Added multi-select functionality in add tags to message dialog box.

## Test Plan

Added


kurund (changes requested):
@rathorevaibhav I think we should switch this to use abstracted autocomplete so that we are consistent with the UI.

Screenshot 2020-09-10 at 11 39 01 PM

src/components/UI/SearchDialogBox/SearchDialogBox.tsx
```diff
@@ -31,6 +32,7 @@ export const SearchDialogBox = (props: any) => {
       <div className={styles.DialogBox}>
         <FormControl fullWidth>
           <AutoComplete
```
DigneshGujarathi, line 34: Is there any reason we're using material autocomplete instead of the AutoComplete component?

DigneshGujarathi (commented):
Looks good

kurund (approved):

PR #456, opened 2020-09-11
Updated organisation name as a name

DigneshGujarathi:


src/containers/OrganisationSettings/OrganisationSettings.test.helper.ts
```diff
@@ -83,6 +79,7 @@ export const LIST_ITEM_MOCKS = [
             providerKey: 'ADD_PROVIDER_API_KEY',
             providerPhone: '917834811114',
             shortcode: 'Glific',
```
kurund, line 81: let's remove shortcode if we are not using.

src/graphql/queries/Organization.ts
```diff
@@ -6,6 +6,7 @@ export const GET_ORGANIZATION = gql`
       organization {
         id
         shortcode
```
kurund, line 8: let's remove shortcode if we are not using it.

kurund (changes requested):

kurund (approved):
Looks good

PR #528, opened 2020-09-23
Check for undefined/null settings and fields

rathorevaibhav:
## Summary

Check for undefined/null settings and fields


DigneshGujarathi (commented):
Looks good

PR #537, opened 2020-09-25
Group Contact removal

rathorevaibhav:
## Summary

When contact is removed from the groups it is directly removed from listing without page reload.

## Test Plan

Updated


kurund (approved):
works perfectly!

Issue #540, opened 2020-09-28
Templates tab should be there?


abhi1203:
@kurund @pankaj-ag 
Any ideas why we can't see the template tab here, even though the session timer shows 0 - which means that the person is opted in?
![image](https://user-images.githubusercontent.com/16714604/94430491-a2606e80-01b1-11eb-90b6-366d7e36fc84.png)

kurund:
@abhi1203 This might be a case where the beneficiary has not opted in.

abhi1203:
Okay and in that case the session timer should also show the block sign instaed of 0. that's where i was confused

rathorevaibhav:
@kurund @pankaj-ag I was able to re-create this case by modifying the database directly. Here's what I have in place for this contact:
1. status: valid
2. provider_status: none

I'm not sure how without modifying DB, these settings can be possible. But thought to post in case it can help to debug.

kurund:
@rathorevaibhav added an additional condition to display blocked icon if provider_status == NONE

PR #547, opened 2020-09-29
Session expiry

kurund:
## Summary

* Fixes as per #540

## Test Plan

* Not needed

### DIFF
--- a/src/components/UI/Timer/Timer.tsx
+++ b/src/components/UI/Timer/Timer.tsx
@@ -7,7 +7,7 @@ import styles from './Timer.module.css';
 export interface TimerProps {
   time: any;
   contactStatus?: string;
-  providerStatus?: string;
+  contactProviderStatus?: string;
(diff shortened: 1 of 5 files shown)


rathorevaibhav (approved):

rathorevaibhav:
@kurund the code changes look good. Just one query from testing:

In case the user is opted out, we show the blocked icon. Should we also hide the chat input?
image

kurund:
@rathorevaibhav Actually for all purposes it's like blocked contact where you can't send a message. However, I feel showing input gives an indication that we can communicate with the user once we receive a message from them. So I feel current behavior might be fine.

@abhi1203 Your thoughts on this?

kurund:
@rathorevaibhav I have made a fix to hide the input when the session has expired and the contact has not opted in.

rathorevaibhav (approved):

PR #553, opened 2020-10-01
Organisation language support

rathorevaibhav:
## Summary

Fixes to get active languages for user

## Test Plan

Updated.

## Note
For resources like tags, speed-sends, templates, profile etc where we have language dropdown, the active languages should show up fine. However, when you change the language, save, go to a different page and then come back, it'll still show the old language from the cache. Upon refresh, it's working fine.

This is currently being worked on on a separate branch.


kurund (approved):
Looks good!

PR #555, opened 2020-10-05
Language bug fixed at all places

rathorevaibhav:
## Summary

Language bug fixed

## Test Plan

Updated


kurund (approved):
tested and working fine.

PR #557, opened 2020-10-05
Prefix in API URL

rathorevaibhav:
## Summary

Prefix option for API url

## Test Plan

Not needed


rathorevaibhav:
@kurund merging this due to urgency. Please review when you get time.

PR #558, opened 2020-10-06
Simulator functionality

rathorevaibhav:
## Summary

Simulator functionality

## Test Plan

Added


src/containers/Chat/Chat.tsx
```diff
@@ -18,6 +18,8 @@ import {
 } from '../../graphql/subscriptions/Tag';
 import { setErrorMessage } from '../../common/notification';
 import { SEARCH_QUERY_VARIABLES } from '../../common/constants';
+import Axios from 'axios';
```
kurund, line 21:
```suggestion
import axios from 'axios';
```

.env.example
```diff
@@ -1,3 +1,4 @@
 REACT_APP_API_PREFIX="api"
 SENTRY_DSN=""
 REACT_APP_GLIFIC_API_PORT=4000
+REACT_APP_SIMULATOR_CONTACT_NUMBER="9876543210"
```
kurund, line 4: Why do we need to define the phone number here? Since it will be used internally should we define it in constant?

src/components/simulator/Simulator.tsx
```diff
@@ -0,0 +1,196 @@
...
+import { useQuery } from '@apollo/client';
+import Draggable from 'react-draggable';
+import DoneAllIcon from '@material-ui/icons/DoneAll';
+import InsertEmoticonIcon from '@material-ui/icons/InsertEmoticon';
+import MicIcon from '@material-ui/icons/Mic';
+import CallIcon from '@material-ui/icons/Call';
+import MoreVertIcon from '@material-ui/icons/MoreVert';
+import VideocamIcon from '@material-ui/icons/Videocam';
+import CameraAltIcon from '@material-ui/icons/CameraAlt';
+import Axios from 'axios';
```
kurund, line 17:
```suggestion
import axios from 'axios';
```

src/components/simulator/Simulator.tsx
```diff
@@ -0,0 +1,196 @@
...
+  showSimulator,
+  setShowSimulator,
+}: SimulatorProps) => {
+  const [inputMessage, setInputMessage] = useState('');
+  const messageRef: any = useRef<HTMLDivElement>();
+
+  let messages = [];
+  let simulatorId = '';
+
+  const { data: allConversations }: any = useQuery(SEARCH_QUERY, {
```
kurund, line 40: maybe here you should get only simulator contact conversations.

kurund (changes requested):

.env.example
rathorevaibhav: this is something we're unsure of what exactly would be the number. So added this flexibility that it can easily changed in Glific installations by others

src/components/simulator/Simulator.tsx
rathorevaibhav: We are filtering the contact from all conversations query because when we send or receive a message from subscription, only this query is updated and the changes are reflected on the screen. If we will not call this query and get only the messages of the particular contact it will be difficult to update on sending and receiving messages in the simulator.

kurund:
@rathorevaibhav Also I am getting some weird errors in the console when I send a message in the simulator.

Screenshot 2020-10-07 at 10 08 27 PM

.env.example
kurund: We are not sending actual messages, so in a way, the number does not matter right?

src/components/simulator/Simulator.tsx
kurund: Ok make sense.

.env.example
rathorevaibhav: the number needs to be tracked by the backend as it is treated as a contact and the messages are stored. I think we can keep it fixed in config.

kurund: Can you remove "REACT_APP_SIMULATOR_CONTACT_NUMBER" from .env.example as it is defined in constant

src/config/index.ts
```diff
@@ -25,3 +25,5 @@ export const USER_SESSION = GLIFIC_API_URL + '/v1/session';
 export const RESET_PASSWORD = GLIFIC_API_URL + '/v1/registration/reset-password';
 export const RENEW_TOKEN = USER_SESSION + '/renew';
 export const FLOW_EDITOR_CONFIGURE_LINK = `${PROTOCOL}//${window.location.host}/automation/configure`;
+export const GUPSHUP_CALLBACK_URL = GLIFIC_BACKEND_URL + '/gupshup';
+export const SIMULATOR_CONTACT = envVariables.REACT_APP_SIMULATOR_CONTACT_NUMBER;
```
kurund, line 29: We should remove this as it is defined in constant

kurund (approved):
All good now 👍

PR #567, opened 2020-10-07
Added description & IsActive field

DigneshGujarathi:
## Summary

* Add a description and "is active" field for credential settings

## Depends on 
* Backend API changes: https://github.com/glific/glific/tree/api-changes

## Test Plan

* Included in the PR


kurund (changes requested):
A couple of issues
1. Is Active checkbox is not shown correctly. We need to hard refresh.
2. Success message persists when you click back to the listing and then come back again to update the creds.

kurund:
@DigneshGujarathi I think we should also redirect to the listing after "Save" action. Earlier we had only settings so it made sense to keep it on the same page.

rathorevaibhav:
Some things I found while testing:

1. Not able to see any description field in the providers
image
2. There is a lot of delay after clicking on save which redirects back to the settings page.
3. It does not save when we try to make isActive true after changing it to false.

DigneshGujarathi:
> 1. Not able to see any description field in the providers
>    image
The description is shown on the setting listing page, not in the form.

src/containers/Form/FormLayout.tsx
```diff
@@ -134,6 +134,7 @@ export const FormLayout: React.SFC<FormLayoutProps> = ({
   const { loading, error } = useQuery(getItemQuery, {
     variables: variables,
     skip: !itemId,
+    fetchPolicy: 'no-cache', // This is required to restore the data after save
```
rathorevaibhav, line 137: instead of not going with cache, can we update the cache?

Basically, if you include the things you want to update in the update query it automatically updates the cache.

rathorevaibhav (approved):

PR #572, opened 2020-10-07
Tag update on collections bug fixed

rathorevaibhav:
## Summary

Newly created tags are now updating in the collections.


kurund (approved):
works well.

PR #575, opened 2020-10-07
Emoji support bug resolved

rathorevaibhav:
## Summary

All whatsapp emoji's are now supported

### DIFF
--- a/src/containers/Chat/ChatMessages/ChatMessage/ChatMessage.module.css
+++ b/src/containers/Chat/ChatMessages/ChatMessage/ChatMessage.module.css
@@ -3,7 +3,7 @@
   position: relative;
   max-width: 100%;
   border-radius: 12px;
-  font-family: 'Heebo', sans-serif;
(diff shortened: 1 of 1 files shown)


src/containers/Chat/ChatMessages/ChatMessage/ChatMessage.module.css
```diff
@@ -3,7 +3,7 @@
   position: relative;
   max-width: 100%;
   border-radius: 12px;
-  font-family: 'Heebo', sans-serif;
+  font-family: 'Heebo', sans-serif, emoji;
```
kurund, line 6: @rathorevaibhav are we including this emoji font somewhere else?

kurund:
@rathorevaibhav Were there any specific emoji's that didn't work? I tried a few and it worked without this patch.

rathorevaibhav:
@kurund the `emoji` font seems like a browser thing. I haven't found anything solid on this but the functionality seems to be working normally across major browsers and on different machines.

PR #577, opened 2020-10-09
Chat input clickable

rathorevaibhav:
## Summary

I made some updates in such a way that there's just one line in the chat input (like WhatsApp). Having a second line would have been confusing and two lines were not in alignment with send and emoji buttons.


kurund (approved):
Good idea!

PR #578, opened 2020-10-09
List sorting

DigneshGujarathi:

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -29,3 +29,14 @@ export const SEARCH_QUERY_VARIABLES = {
     limit: 50,
   },
 };
+
+export const SET_VARIABLES = (filter: any, limit: any, offset: any, order: string) => {
+  return {
+    filter: filter,
+    opts: {
+      limit: limit,
+      offset: offset,
+      order: order,
+    },
+  };
+};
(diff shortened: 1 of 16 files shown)


src/common/constants.ts
```diff
@@ -29,3 +29,14 @@ export const SEARCH_QUERY_VARIABLES = {
     limit: 50,
   },
 };
+
+export const SET_VARIABLES = (filter: any, limit: any, offset: any, order: string) => {
```
kurund, line 33: @DigneshGujarathi may be set defaults to the params so that we can further optimize the rest of the code.

src/containers/Chat/ChatMessages/ContactBar/ContactBar.tsx
```diff
@@ -50,10 +50,14 @@ export const ContactBar: React.SFC<ContactBarProps> = (props) => {
   const [showBlockDialog, setShowBlockDialog] = useState(false);
 
   // get group list
-  const [getGroups, { data: groupsData }] = useLazyQuery(GET_GROUPS);
+  const [getGroups, { data: groupsData }] = useLazyQuery(GET_GROUPS, {
+    variables: SET_VARIABLES({}, null, 0, 'ASC'),
```
kurund, line 54: here we can directly call setVariables()

kurund (changes requested):

kurund (approved):
Works great!

PR #601, opened 2020-10-12
Added multi search

DigneshGujarathi:


rathorevaibhav:
@DigneshGujarathi When I search and click on the message and if the conversation is already not loaded it gets stuck in the loading state. Not sure if I'm doing it right. I'll check again.

rathorevaibhav:
@DigneshGujarathi some testing feedbacks:

1. When searching for a message and clicking on the contact which is initially not loaded in the search query, it gets stuck in the loading state

Screenshot from 2020-10-21 10-36-08
Screenshot from 2020-10-21 10-36-34
img


2. The order of the search should be `contact > messages > tags` according to the sketch files. Currently its `contact > tags > messages`

3. When the result is null the list area should not be blank. It can have a message like this- "no contact found".
image

rathorevaibhav:
@DigneshGujarathi  One final thing I found on testing. When not searching for anything and clicking on any contact, it duplicates that contact in the list:

Screenshot from 2020-10-22 12-30-30

rathorevaibhav (approved):

PR #606, opened 2020-10-13
Load more contacts when scrolling to the bottom of the contacts

rathorevaibhav:
## Summary

Ability to load more contacts when we reach the bottom of the contacts list

## Test Plan

Added


kurund (approved):
Looks good!

PR #607, opened 2020-10-13
Load more messages on scrolling up the chat

rathorevaibhav:
## Summary

Now we can view all previous conversations when we scroll to the top of the messages

## Test Plan

Added


kurund (approved):
works great!

PR #617, opened 2020-10-16
Support for adding bulk contacts to group

rathorevaibhav:
## Summary

We can add bulk contacts to a group from the group list screen


kurund (changes requested):
@rathorevaibhav Mostly everything is working as expected. However, I am getting this weird behavior when I search for contact and then select the contact. The form just breaks after that.
Screenshot 2020-10-20 at 6 56 36 PM

rathorevaibhav:
hey @kurund I tried a few combinations but couldn't replicate it. can you share some more details or maybe the exact steps you're doing?

kurund:
We already discussed, just adding it here for future reference. It happens when you try to search for a contact that is not part of the preloaded contacts.

kurund (approved):
Looks good and working as expected now.

Issue #620, opened 2020-10-19
Websocket connection never stops reconnection attempt incase of failure


kurund:
**Describe the bug**
* In case of failure in WebSocket connection, it tries to auto-reconnect. However, it continues to reconnect infinitely until the connection is established. 

**Expected behavior**
* It should stop reconnection attempts after certain attempts

PR #625, opened 2020-10-19
let's stop ws retry

kurund:
## Summary

* Fixes for #620

## Test Plan

* Not needed

### DIFF
--- a/src/config/absinthe.ts
+++ b/src/config/absinthe.ts
@@ -18,10 +18,18 @@ const socketConnection = new PhoenixSocket.Socket(SOCKET, {
   },
 });
 
-// watch for websocket error / close event
-socketConnection.onClose((event: any) => {
-  // don't retry as if connection is closed due to some failures and cancel automatic reconnect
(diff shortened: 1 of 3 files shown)


src/config/absinthe.ts
```diff
@@ -18,5 +18,19 @@ const socketConnection = new PhoenixSocket.Socket(SOCKET, {
   },
 });
 
+// we should try to reconnect ws connection only finite (5) times and then abort and prevent
+// unnecessary load on the server
+// watch for websocket error event using onError
+const WEB_SOCKET_RETRY_ATTEMPTS = 5;
```
rathorevaibhav, line 24: I think this should be in constants file?

rathorevaibhav (commented):
Just one comment. Rest looking good.

src/config/absinthe.ts
kurund: I was thinking about it too, not sure why I ruled against it :) Let me make the change.

rathorevaibhav (approved):

PR #631, opened 2020-10-21
background color updated for create collection button

rathorevaibhav:
## Summary

The create collection and cancel button now have a white background instead of transparent.

## Test Plan

Not needed


kurund (changes requested):
We should fix the update button.

Screenshot 2020-10-22 at 10 01 11 AM

rathorevaibhav:
@kurund this is ready for review

kurund (approved):
Looks good!

PR #641, opened 2020-10-26
Added warning popup on automation configure page when navigating away without saving

rathorevaibhav:
## Summary

Added warning popup for the user to not leave without saving changes.

## Test cases
https://github.com/glific/cypress-testing/pull/15


kurund (approved):
working as expected

PR #642, opened 2020-10-26
Updated UI for chat blast feature

rathorevaibhav:
## Summary

UI updated according to the design files

## Test Plan

Updated


kurund (approved):
Works perfectly.

Minor UI glitch that can be fixed in separate PR. The width of the popup needs to same as the chat input container.

Screenshot 2020-10-26 at 8 10 56 PM

PR #653, opened 2020-10-28
Tab title will update according to the automation page

rathorevaibhav:
## Summary

The automation title will be appended to the title of the website. Also added environment variables for:
1. App name
2. Page title


.env.example
```diff
@@ -1,3 +1,5 @@
 REACT_APP_API_PREFIX=""
 SENTRY_DSN=""
 REACT_APP_GLIFIC_API_PORT=4000
+REACT_APP_APPLICATION_NAME=""
```
kurund, line 4: @rathorevaibhav why do we need an app name?

.env.example
```diff
@@ -1,3 +1,5 @@
...
+REACT_APP_PAGE_TITLE=""
```
kurund, line 5: Why do we need to define title as env variable?

kurund (changes requested):

.env.example
rathorevaibhav: hey @kurund I made these fields editable from env so that if needed, SaaS providers who extend Glific can rename it.

For example, like Tides, we can have App name set to "Tides" instead of "Glific". It's just so that we provide flexibility to extend the framework. If there's nothing set, the default value will come in which is "Glific".

rathorevaibhav:
@kurund this is ready for review again

kurund (approved):
Looks good!

PR #658, opened 2020-10-30
Subscription reloaded

kurund:
## Summary

* Abstract chat subscriptions and move to separate component
* Move subscriptions to top-level so that it can listen to other events like the message to a group

## Test Plan

* Not needed


rathorevaibhav:
I can't log in. The login page redirects to the logout page and vice versa in a loop. @kurund can you check if it's happening for you too?

kurund:
@rathorevaibhav working as expected for me.

rathorevaibhav:
hey @kurund I'm still getting the issue with the login and it keeps reloading on loop.

kurund:
@rathorevaibhav Fixed the issue related to logout

PR #673, opened 2020-11-03
Block contact loading bug removed

rathorevaibhav:
## Summary
After blocking a contact the loading window is no more there. It redirects back to the chat route.


kurund (approved):
Looks good!

PR #674, opened 2020-11-04
Clear apollo cache on logout

mdshamoon:
## Summary

Apollo cache will be cleared once the user logout.

### DIFF
--- a/src/containers/Auth/Logout/Logout.tsx
+++ b/src/containers/Auth/Logout/Logout.tsx
@@ -25,7 +25,7 @@ export const Logout: React.SFC<LogoutProps> = () => {
   resetRole();
   
   // clear apollo cache
-  client.resetStore();
+  client.clearStore();
 
   return <Redirect to="/login" />;
 };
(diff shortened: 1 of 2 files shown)


src/containers/Auth/Logout/Logout.tsx
```diff
@@ -21,8 +23,9 @@ export const Logout: React.SFC<LogoutProps> = () => {
 
   // clear role & access permissions
   resetRole();
-
-  // TODOS: We should clear apollo cache
+  
+  // clear apollo cache
+  client.resetStore();
```
kurund, line 28: @mdshamoon I think we should use clearStore() here as we don't want to fetch the queries again.  Check https://www.apollographql.com/docs/react/api/core/ApolloClient/#ApolloClient.resetStore

Thoughts?

kurund (changes requested):

src/containers/Auth/Logout/Logout.tsx
mdshamoon: @kurund Yes, I believe clearStore() makes more sense as we don't need to refetch the queries again. Let me update the function.

PR #677, opened 2020-11-04
Toast service added

rathorevaibhav:
## Summary

Added service for toast message that is called on all the authenticated routes.
Also removed redundant code.

## Test Plan

Updated


kurund (approved):
Excellent cleanup @rathorevaibhav 🎉

PR #719, opened 2020-11-19
Added support for whatsapp bold+italic formatting

rathorevaibhav:
## Summary

Added support for WhatsApp formatting.


kurund (changes requested):
@rathorevaibhav bold+italic is not working. I also noticed that formatting is not shown correctly in the simulator.
Screenshot 2020-11-20 at 9 25 05 PM

rathorevaibhav:
@kurund this is ready for review again.

Just one clarification, this PR does not target the keyboard shortcuts. I'll have them ready in a separate PR as it requires some digging into draftjs editor support.

This PR only contains formatting corrections, bold+italics, in simulator window and chat window if we're sending typed messages with asterisks(*) and underscores(_).

kurund (approved):
working fine when manually formatted.

PR #721, opened 2020-11-20
Added subscription for wallet balance

DigneshGujarathi:


kurund (changes requested):
@DigneshGujarathi Overall functionality and the display is fine. However, realtime update using subscriptions is not working correctly. I need to reload the page in order to see the correct wallet balance.

DigneshGujarathi:
> @DigneshGujarathi Overall functionality and the display is fine. However, realtime update using subscriptions is not working correctly. I need to reload the page in order to see the correct wallet balance.

The subscription duration is set for every half hour. So this refreshes the balance every half-hour cycle.
We may decrease the interval time as per the discussion with @pankaj-ag if necessary.

kurund (approved):
Ok, working as expected. Changed the backend cron to 1 min and tested.

Issue #717, opened 2020-11-19
User permissions in Glific


rathorevaibhav:
**Describe the task**
All the targeted items listed in "User roles and permissioning" section:
https://docs.google.com/document/d/1Q7ozYAHl1wcVQlXcg3iZjQZPHaUQJzQCYHTLaajncrs/edit#heading=h.uy9zxlqp6ucl

PR #723, opened 2020-11-20
Simulator access based on user role

rathorevaibhav:
## Summary

A staff member will not be given access to the simulator.


kurund (approved):
Looks good!

PR #724, opened 2020-11-20
Roles help modal added

rathorevaibhav:
## Summary

Added a modal to specify the permissions associated with each role on the add/edit staff page.


kurund (approved):
works well!

Issue #712, opened 2020-11-17
Number masking


abhi1203:
At the top and in the right panel
![image](https://user-images.githubusercontent.com/16714604/99357578-24c3ff80-28d2-11eb-9f8c-ab99131dff1f.png)

while searching
![image](https://user-images.githubusercontent.com/16714604/99357668-4b823600-28d2-11eb-8f05-fdd3b87aea61.png)

in blocked contacts
![image](https://user-images.githubusercontent.com/16714604/99361875-c64e4f80-28d8-11eb-9c8f-59ab5fd9dfdc.png)

in groups
![image](https://user-images.githubusercontent.com/16714604/99362121-14635300-28d9-11eb-9cfc-84d4a01bf2dd.png)

in contact profile
view number icons when it's hidden then the hide icon when the number is showing
![image](https://user-images.githubusercontent.com/16714604/99362420-7b810780-28d9-11eb-8983-90aa2e27e536.png)

![image](https://user-images.githubusercontent.com/16714604/99362596-b125f080-28d9-11eb-90c1-be270779866e.png)

Let me know in case im missing any place.

For staff, the numbers will always be hidden.

PR #727, opened 2020-11-20
let's mask the phone number

kurund:
## Summary

* Introduce phone number masking. (#712)

## Test Plan

* Included in the PR


rathorevaibhav:
hey @kurund sorry about the late comment. All things looking good. Just one minor thing. On the contact description page, when we click on the eye icon it slightly shifts to the left as the spacing increases when numbers are fully displayed vs when they're masked. Do you think this is something we should fix or can leave it be?

kurund:
@rathorevaibhav I will fix it.

kurund:
@rathorevaibhav Ready for your review again. Also fixed the phone value for a staff role user and added a tooltip.

rathorevaibhav (approved):

PR #728, opened 2020-11-21
Added restriction checkbox for staff role

rathorevaibhav:
## Summary

If the user role will be staff then a checkbox will be added on the staff edit screen to restrict the access of the staff.


kurund (approved):
tested and works perfectly!

PR #731, opened 2020-11-22
logout fixes

kurund:
## Summary

* In some weird situations after logout, the page keeps looping between login and logout

## Test Plan

* Not needed


DigneshGujarathi (commented):
Looks good

PR #735, opened 2020-11-23
Admin changing their role should log them out

rathorevaibhav:
## Summary

If an admin changes their role, the system logs them out.


kurund (approved):
Looks good!

PR #736, opened 2020-11-23
Only admin can manage user roles

rathorevaibhav:
## Summary

Removed manager and staff access to staff management


kurund (changes requested):
@rathorevaibhav I think Manager needs to have access to staff listing. However, she should not be able to update the role of the staff. Maybe we can make the role field read-only. @abhi1203 Any thoughts on this?

kurund (approved):
Looks good

Issue #714, opened 2020-11-18
In simulator start automation is available but didn't trigger


abhi1203:
**Describe the bug**
For simulator, the session had timed out. But the start automation option is available unlike for regular contacts. However, starting the flow didn't trigger the message. it got triggered when message was sent from the simulator phone first.

**To Reproduce**
1. wait for timeout in simulator
2. start automation flow
3. select a flow(Activity) and start
4. See error that the flow doesn't start

[Screen Recording 2020-11-18 at 2.28.25 PM.zip](https://github.com/glific/glific-frontend/files/5558988/Screen.Recording.2020-11-18.at.2.28.25.PM.zip)

in the image: automation worked when hi was sent from simulator first
<img width="1440" alt="Screenshot 2020-11-18 at 2 29 07 PM" src="https://user-images.githubusercontent.com/16714604/99509477-d125e480-29ab-11eb-9e96-f9188df9dc87.png">

PR #740, opened 2020-11-24
Simulator timeout bug resolved

rathorevaibhav:
## Summary

Fixes for #714


kurund:
@rathorevaibhav as discussed we also need to handle this condition for chat input too.

rathorevaibhav:
@kurund done with the feedback. Please have a look at this again.

kurund (approved):
Looks good!

PR #741, opened 2020-11-24
sticker support added

rathorevaibhav:
## Summary

Added support for Whatsapp stickers. This needs https://github.com/glific/glific/issues/689 to be implemented at backend.

## Test Plan

Added


kurund (approved):
Looks good, hence merging this.

@rathorevaibhav we should also remove chat background. (green)

Screenshot 2020-11-27 at 7 35 27 PM

This could be done in separate PR

PR #748, opened 2020-11-25
route fixes

kurund:
## Summary

* Clean up the route implementation

## Test Plan

* Not needed

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -29,20 +29,21 @@ const App = () => {
 
   let routes;
 
-  if (authenticated !== undefined)
+  if (authenticated !== undefined) {
     if (authenticated) {
       routes = <AuthenticatedRoute />;
     } else {
       routes = <UnauthenticatedRoute />;
     }
 
-  // For logout acion, we don't need to check if the user is logged in or not. Hence, adding it at top level
-  routes = (
-    <Switch>
-      <Route path="/logout" exact component={Logout} />
-      {routes}
-    </Switch>
-  );
+    // For logout acion, we don't need to check if the user is logged in or not. Hence, adding it at top level
+    routes = (
+      <Switch>
+        <Route path="/logout" exact component={Logout} />
+        {routes}
+      </Switch>
+    );
+  }
 
   return (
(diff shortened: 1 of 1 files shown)


kurund:
@rathorevaibhav sorry forgot to comment earlier, tested few scenarios pending scenarios, and is working as expected. Ready for your review.

src/App.tsx
```diff
@@ -29,20 +29,21 @@ const App = () => {
...
     }
 
-  // For logout acion, we don't need to check if the user is logged in or not. Hence, adding it at top level
-  routes = (
-    <Switch>
-      <Route path="/logout" exact component={Logout} />
-      {routes}
-    </Switch>
-  );
+    // For logout acion, we don't need to check if the user is logged in or not. Hence, adding it at top level
```
rathorevaibhav, line 39:
```suggestion
    // For logout action, we don't need to check if the user is logged in or not. Hence, adding it at top level
```

src/App.tsx
```diff
@@ -29,20 +29,21 @@ const App = () => {
 
   let routes;
 
-  if (authenticated !== undefined)
+  if (authenticated !== undefined) {
     if (authenticated) {
       routes = <AuthenticatedRoute />;
     } else {
       routes = <UnauthenticatedRoute />;
     }
```
rathorevaibhav, line 37: Very minor thing. Do you think something like this is valid?
```suggestion
  if (authenticated !== undefined && authenticated) {
    routes = <AuthenticatedRoute />;
  } else {
      routes = <UnauthenticatedRoute />;
  }
```

rathorevaibhav (commented):
one question and one minor feedback. @kurund have a look

src/App.tsx
kurund: @rathorevaibhav Actually we don't want to assign any routes, hence the current logic. In the above scenario, we will set UnAuthenticated routes.

rathorevaibhav: Sounds good then.

rathorevaibhav (approved):

PR #750, opened 2020-11-26
Added support for sending media attachments on chat screen

rathorevaibhav:
## Summary

Added an attachment icon next to the emoji icon.
On clicking the attachment icon a dialog box opens up where the user can select the type and URL of the message they want to send.

## Test Plan

Added


kurund (changes requested):
Few issues:
* Click attachment, pop up opens. Hit save without entering anything and then click on Send. It throws below error.
Screenshot 2020-11-27 at 7 40 02 PM
* Stickers not working ( Note I also tried merging master so that we have sticker work)

kurund:
Another issue:

If I do below
Screenshot 2020-11-27 at 8 00 36 PM

and send then I get this weird error
Screenshot 2020-11-27 at 8 01 05 PM

This might be because the backend is throwing errors. So we might need to handle this use case.

rathorevaibhav:
@kurund done with the feedbacks. Please have a look

kurund (approved):
The above-mentioned issues are resolved. I guess ready to ship this functionality 👍

Issue #759, opened 2020-11-27
Font properties on settings page


abhi1203:
![image](https://user-images.githubusercontent.com/16714604/100412473-d5808a80-309a-11eb-8466-39e269368874.png)

- [ ] Please check the spacing, between heading and text.

- [ ] the font size and color

reference:
![image](https://user-images.githubusercontent.com/16714604/100412543-0496fc00-309b-11eb-8b4d-87680a14e888.png)

PR #766, opened 2020-11-30
Fixed font and style settings on settings page

rathorevaibhav:
## Summary

Fixes for #759


kurund (approved):
Looks great. Just made a minor CSS adjustment so that it looks good in higher resolution too.

Issue #749, opened 2020-11-26
date tooltip doesnt show for messages with image 


abhi1203:
![image](https://user-images.githubusercontent.com/16714604/100311925-eff13000-2fd6-11eb-8f82-32f06570486e.png)

Also can we show the tooltip by hovering over any area of the message, currently it shows only when hovered on text.

PR #770, opened 2020-12-01
Tooltip added on media

rathorevaibhav:
## Summary

Fixes for #749


src/components/UI/Layout/Navigation/SideDrawer/SideDrawer.tsx
```diff
@@ -243,34 +243,38 @@ export const SideDrawer: React.SFC<SideDrawerProps> = () => {
...
-                          location.pathname
-                        )
-                          ? ActiveStaffIcon
-                          : InactiveStaffIcon
-                      }
-                      className={styles.StaffIcon}
-                      alt="staff icon"
-                    />
-                  </IconButton>
+                  <span>
```
kurund, line 246: @rathorevaibhav SPAN is an inline element. This may work, but may not be the best practice.

kurund (changes requested):

rathorevaibhav:
@kurund ready for review again

kurund (approved):
Looks good!

PR #789, opened 2020-12-03
Automation autosave when moving out of floweditor removed 

rathorevaibhav:
## Summary

When the flow editor component unmounts it will stop all the repetitive autosave calls.


DigneshGujarathi (commented):
Looks good

PR #793, opened 2020-12-04
Search API multiple calls on sending a message fixed

rathorevaibhav:
## Summary

Chats will now be retrieved from the cache instead of the network.

## Test Plan

Updated.
- Some test cases need research. I'll target them separately.


DigneshGujarathi (commented):
Looks good

PR #794, opened 2020-12-06
Updated template form

DigneshGujarathi:


rathorevaibhav:
A few things on testing here:

1. If we save template without adding the URL it gives the following error
image

2. Alignment not proper in english and hindi language
image

3. If we click on the show all languages button of the first item and then the second item, it closes the all languages list for the first item but the second item is not opened

4. When language is changed in edit template the title placeholder overlaps with title, also attachment type changes to null but attachment URL remains the same

src/containers/Form/FormLayout.tsx
```diff
@@ -50,6 +51,7 @@ export interface FormLayoutProps {
   title?: string;
   getLanguageId?: any;
   backLinkButton?: any;
+  IsAttachment?: boolean;
```
rathorevaibhav, line 54: should be
```suggestion
  isAttachment?: boolean;
```

PR #801, opened 2020-12-08
Added remove attachment option

rathorevaibhav:
## Summary

Attachment can now be removed after adding.


DigneshGujarathi (commented):
Looks good

PR #805, opened 2020-12-10
Update chat templates

DigneshGujarathi:


src/containers/Template/Form/Template.tsx
```diff
@@ -21,12 +24,15 @@ const FormSchema = Yup.object().shape({
...
+  language: Yup.object().nullable().required('Language is required.'),
 });
 
+const pattern = /[^{}]+(?=})/g;
+
 const dialogMessage = ' It will stop showing when you are drafting a customized message.';
 
 const defaultTypeAttribute = {
-  type: 'TEXT',
+  // type: 'TEXT',
```
rathorevaibhav, line 35: is this needed or can be removed easily?

src/containers/Form/FormLayout.tsx
```diff
@@ -216,6 +222,38 @@ export const FormLayout: React.SFC<FormLayoutProps> = ({
     return null;
   }
 
+  const getMediaId = (payload: any) => {
```
rathorevaibhav, line 225: I think this is not something we should include in FormLayout.tsx. Maybe in parent or maybe we create another (generic) wrapper around FormLayout that deals with media. Let me know your thoughts.

DigneshGujarathi: I have fixed this, please review.

PR #810, opened 2020-12-14
Fixed error in the build process

rathorevaibhav:
## Summary

Used a stable version for buildpack. 

Basically the buildpack we were using to build the app updated their Nginx version which caused some issues. So now we're using a stable version of the buildpack and its should work fine.


DigneshGujarathi (commented):
Looks good

PR #825, opened 2020-12-16
Added variable popup

DigneshGujarathi:


mdshamoon:
Testing feedbacks:

1. The default height of chat Input is increased. This is affecting the alignment of icons and the placeholder.

image

2. If a template has no variable, it's treating it like a normal message in the chat input field and we are able to edit that.

image

mdshamoon:
Testing feedback:

Account update template has 3 variables but in the dialog box, it's showing only one.

image

image

mdshamoon:
Looks good.

PR #833, opened 2020-12-18
Scroll position retained when clicking on a new contact

mdshamoon:
## Summary

Scroll position retained when we click on a new contact


DigneshGujarathi (commented):
Looks good

PR #834, opened 2020-12-18
Reverted translations for HSM

DigneshGujarathi:


src/containers/Form/FormLayout.tsx
```diff
@@ -98,6 +100,8 @@ export const FormLayout: React.SFC<FormLayoutProps> = ({
   const [action, setAction] = useState(false);
   const [link, setLink] = useState(undefined);
   const [deleted, setDeleted] = useState(false);
+  const [SaveClick, onSaveClick] = useState(false);
```
mdshamoon, line 103:
```suggestion
  const [saveClick, onSaveClick] = useState(false);
```

mdshamoon:
One minor feedback

image

The sample message is still editable on the edit screen.

mdshamoon:
Rest functionality is working fine.

DigneshGujarathi:
> One minor feedback
> 
> image
> 
> The sample message is still editable on the edit screen.

@mdshamoon It looks like an issue with the EmojiInput component, Can we handle it as a separate issue?

mdshamoon:
> @mdshamoon It looks like an issue with the EmojiInput component, Can we handle it as a separate issue?

Sure. Let's track it in a separate issue.

PR #837, opened 2020-12-21
Send OTP twice bug resolved

mdshamoon:
## Summary

Here is a reference to the solution: https://stackoverflow.com/questions/61557815/react-formik-onsubmit-async-called-twice#comment108889373_61557815

### DIFF
--- a/src/containers/Auth/Auth.module.css
+++ b/src/containers/Auth/Auth.module.css
@@ -212,3 +212,7 @@
 .OTPBox fieldset {
   border-color: #93a29b !important;
 }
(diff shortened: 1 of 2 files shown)


kurund (approved):
@mdshamoon @DigneshGujarathi  Removing "type=submit" results in old behavior where we cannot submit the login form using the 'enter' key. I have added an additional fix to handle that case. Can you please verify both the fixes at your end?

mdshamoon:
Working fine for me 👍. It's handling both cases.

PR #841, opened 2020-12-22
Sorting stays the same even after refresh

mdshamoon:
## Summary

List sorting stays the same even after refresh or page switching.

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -64,12 +64,14 @@ export const MEDIA_MESSAGE_TYPES = ['IMAGE', 'AUDIO', 'VIDEO', 'DOCUMENT', 'STIC
 export const setColumnToBackendTerms: any = (
   listName: string,
   columnName: string,
-  check: boolean = false
+  checkLastSort: boolean = false
 ) => {
   const backendTerms: any = {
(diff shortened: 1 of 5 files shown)


src/common/constants.ts
```diff
@@ -78,5 +84,16 @@ export const setColumnToBackendTerms: any = (listName: string, columnName: strin
     backendTerms.DESCRIPTION = 'label';
   }
 
-  return backendTerms[columnName];
+  let lastSort = null;
```
kurund, line 87: @mdshamoon We should only define constants here. We should move this logic to List.tsx and define a new method to handle this.

src/services/ListService.ts
```diff
@@ -0,0 +1,65 @@
+export const setListSession = (sorting: string) => {
```
kurund, line 1: @mdshamoon Can you please add documentation for each method here. What they do etc.

kurund (changes requested):

kurund:
Overall functionality is working as expected.

kurund (approved):
Looks good!  Did minor clean up. We can merge once CI passes 👍

PR #845, opened 2020-12-23
has location and remove contact group function

pankaj-ag:
- Updated package JSON with the new version
- Remove contact from group


DigneshGujarathi (approved):
Looks good

PR #863, opened 2020-12-29
Fixed remove all data from group

DigneshGujarathi:


src/components/UI/Form/AutoComplete/AutoComplete.tsx
```diff
@@ -100,6 +100,9 @@ export const AutoComplete: React.SFC<AutocompleteProps> = ({
               setSearchTerm('');
               onChange('');
             }
+            if (asyncSearch && value.length === 0) {
```
mdshamoon, line 103:
```suggestion
          else if (asyncSearch && value.length === 0) {
```
We don't need to check both conditions every time. Either one will be true at a time

PR #874, opened 2020-12-30
Removed prefix

DigneshGujarathi:


src/config/index.ts
```diff
@@ -5,9 +5,15 @@ const appName = envVariables.REACT_APP_APPLICATION_NAME;
...
-  : window.location.hostname;
+// added support for localhost, need to update it after backend fixes
+let HOSTNAME;
+if (window.location.hostname === 'localhost') {
+  HOSTNAME = 'localhost';
+} else if (API_PREFIX) {
+  HOSTNAME = `${API_PREFIX}.${window.location.hostname}`;
+} else {
+  HOSTNAME = window.location.hostname;
+}
```
mdshamoon, line 16: I don't think these changes are needed. It will work for localhost if you don't add the API_PREFIX in the env.

mdshamoon, line 16:
```suggestion
let HOSTNAME = window.location.hostname;
if (API_PREFIX && HOSTNAME !== 'localhost') {
  HOSTNAME =`${API_PREFIX}.${HOSTNAME}`;
} 
```

PR #914, opened 2021-01-13
Webhook status field update

mdshamoon:
## Summary

Added status field for webhook logs


DigneshGujarathi (commented):
Currently, if we sort by status list gets empty for now can we disable sort by status to avoid any confusion.

mdshamoon:
> Currently, if we sort by status list gets empty for now can we disable sort by status to avoid any confusion.

Sure. I have disabled the sort for now. Can you have a look again?

PR #937, opened 2021-01-20
Added validation for keyword

DigneshGujarathi:

### DIFF
--- a/src/containers/Flow/Flow.tsx
+++ b/src/containers/Flow/Flow.tsx
@@ -20,8 +20,11 @@ export interface FlowProps {
   match: any;
 }
 
+const regex = /^[a-z]+(,[a-z]+)*$/g;
+
 const FormSchema = Yup.object().shape({
   name: Yup.string().required('Name is required.'),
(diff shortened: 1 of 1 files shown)


src/containers/Flow/Flow.tsx
```diff
@@ -20,8 +20,11 @@ export interface FlowProps {
   match: any;
 }
 
+const regex = /^[a-z]+(,[a-z]+)*$/g;
+
 const FormSchema = Yup.object().shape({
   name: Yup.string().required('Name is required.'),
+  keywords: Yup.string().matches(regex, 'Only lowercase are allowed.'),
```
mdshamoon, line 27:
```suggestion
  keywords: Yup.string().matches(regex, 'Only lowercase are allowed.'),
```
Can we have a generic error message as currently if we press space then also we get the same error.

DigneshGujarathi: Space is also not a valid input for this, it only allows comma separated lowercase keywords. We also have helper text for this.

mdshamoon:
Looks good

PR #943, opened 2021-01-21
Search fix

mdshamoon:
## Summary

Added fix for search when the contact does not have any message


DigneshGujarathi (commented):
Looks good

PR #947, opened 2021-01-21
Resolved jump to latest button & error popup

DigneshGujarathi:


mdshamoon:
Added one minor fix for setting the jump to latest button at the center. Rest looks good.

Issue #948, opened 2021-01-22
401 error message


mohitsharma-22:
**Describe the bug**
Users get an error message "**A network error has occurred, Response not successful - Received status code 401**" when they come back to Glific after keep it idle for some time.

**To Reproduce**

lobo Today at 10:00
hey @Glific Frontend
steps to reproduce
get system yp and running
log in to frontend
do a few things
kill backend
rm -rf Mnesia*
mix phx.server
try doing stuff from frontend now
get 401 error
and if u r in simulator, it just eats the messages and nothing happens

**Expected behavior**
It should redirect to the login page.

**Screenshots**
![401 error messgae](https://user-images.githubusercontent.com/16541548/105459650-9c3a5480-5cb0-11eb-9426-8a234890368c.jpeg)

abhi1203:
@DigneshGujarathi should there be a message on the login screen, why they were logged out?

DigneshGujarathi:
> @DigneshGujarathi should there be a message on the login screen, why they were logged out?

If NGOs face any problems with this, may we think about it? To do this, we need to identify the cause for logouts, such as the User action or the Network Problem.

PR #951, opened 2021-01-24
Handling failed authentication

mdshamoon:
## Summary

Fix for #948


src/config/apolloclient.ts
```diff
@@ -65,7 +65,14 @@ const gqlClient = () => {
 
     if (networkError) {
       // eslint-disable-next-line
```
DigneshGujarathi, line 67: Do we need it now?

mdshamoon: No. I will remove it.

DigneshGujarathi (commented):
Looks good

PR #957, opened 2021-01-25
Added validation for attachment URL

DigneshGujarathi:


mdshamoon:
Are we also planning to add a loader because sometimes there is a delay in getting the response back and the user does not know what to do in that case?

mdshamoon:
After clicking on OK I get the error URL is required even if the URL is present.
image

DigneshGujarathi:
> Are we also planning to add a loader because sometimes there is a delay in getting the response back and the user does not know what to do in that case?

Checking with the backend on this

mdshamoon:
Looks good.

PR #960, opened 2021-01-27
Added tooltip for session timer

DigneshGujarathi:


src/components/UI/Timer/Timer.tsx
```diff
@@ -20,10 +21,22 @@ export const Timer: React.FC<TimerProps> = (props: TimerProps) => {
...
+      >
+        <ContactOptOutIcon />
+      </Tooltip>
+    );
   }
 
   let timerStyle = styles.TimerNormal;
+  let tooltipStyle = styles.TimerNormalTooltip;
+  let tooltip =
+    'Session window is open to message this contact. Learn more about the WhatsApp session window here.';
```
mdshamoon, line 39:
```suggestion
   Learn more about the WhatsApp session window here.';
```
Do we need this line since there is no link provided?

mdshamoon, line 39:
```suggestion
    'Session window is open to message this contact. Learn more about the WhatsApp session window here.';
```
Do we need the last line since there is no link provided?

mdshamoon, line 39: Do we need the last line since there is no link provided

DigneshGujarathi: I have updated this with the link

mdshamoon:
The color of the session timer has changed to grey. It should be purple when the time remaining is 22
image

mdshamoon:
Looks good👍

PR #968, opened 2021-02-01
Code cleanup

DigneshGujarathi:

### DIFF
--- a/src/containers/Chat/ChatMessages/ChatMessages.tsx
+++ b/src/containers/Chat/ChatMessages/ChatMessages.tsx
@@ -239,16 +239,21 @@ export const ChatMessages: React.SFC<ChatMessagesProps> = ({
   // use contact id to filter if it is passed via url, else use the first conversation
   let conversationInfo: any = {};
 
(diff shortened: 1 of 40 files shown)


src/containers/Chat/ChatMessages/ChatMessages.tsx
```diff
@@ -158,7 +158,18 @@ export const ChatMessages: React.SFC<ChatMessagesProps> = ({
     refetchQueries: [{ query: SEARCH_QUERY, variables: SEARCH_QUERY_VARIABLES }],
   });
 
-  // this function is called when the message is sent
+  const updatePayload = (payload: any, selectedTemplate: any, variableParam: any) => {
+    // add additional param for template
+    if (selectedTemplate) {
+      /* eslint-disable no-param-reassign */
+      payload.isHsm = selectedTemplate.isHsm;
```
mdshamoon, line 165: We should not update the function parameters. Instead, we can create another variable that can be updated.

mdshamoon:
Looks good.

Issue #971, opened 2021-02-02
Not able to add or remove contacts from collection screen


mdshamoon:
**Describe the bug**
Not able to add or remove contacts from collections

**To Reproduce**
1. Go to Collections
2. Click on add contacts to the collection icon
3. Try to remove or add a contact
4. It will not work

**Expected behavior**
We should be able to add contacts to collections

PR #972, opened 2021-02-02
Bug fixed for add/remove contacts from collection

mdshamoon:
## Summary

Fix for #971

### DIFF
--- a/src/containers/Collection/CollectionList/CollectionList.tsx
+++ b/src/containers/Collection/CollectionList/CollectionList.tsx
@@ -91,8 +91,8 @@ export const CollectionList: React.SFC<CollectionListProps> = () => {
   );
   const [updateCollectionContacts] = useMutation(UPDATE_COLLECTION_CONTACTS, {
     onCompleted: (data) => {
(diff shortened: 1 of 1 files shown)


src/containers/Collection/CollectionList/CollectionList.tsx
```diff
@@ -91,8 +91,8 @@ export const CollectionList: React.SFC<CollectionListProps> = () => {
   );
   const [updateCollectionContacts] = useMutation(UPDATE_COLLECTION_CONTACTS, {
     onCompleted: (data) => {
-      const { numberDeleted, collectionContacts } = data.updateCollectionContacts;
-      const numberAdded = collectionContacts.length;
+      const { numberDeleted, groupContacts } = data.updateGroupContacts;
```
DigneshGujarathi, line 94: We can keep const groupContacts as collectionContacts

mdshamoon: We are getting from the backend in this format only.
image

DigneshGujarathi: Got it, will need to check with the backend for the migration from group to collection

PR #975, opened 2021-02-02
Fixed icon size for search

mdshamoon:
## Summary

On some screens, the search icon was larger than its actual size. So, added a fixed size for the icon


DigneshGujarathi (commented):
Looks good

PR #985, opened 2021-02-04
Updated queries with changed structure

mdshamoon:
## Summary

Updated queries with changed structure


DigneshGujarathi (commented):
Looks good

PR #988, opened 2021-02-05
Fixes for flow keywords

mdshamoon:
## Summary

- Removed restriction for having spaces in between multiple comma-separated flow keywords.
- Allow support for all languages

### DIFF
--- a/src/common/notification.ts
+++ b/src/common/notification.ts
@@ -24,7 +24,7 @@ export const setErrorMessage = (client: any, error: any) => {
     };
 
     if ('key' in error) {
-      errorMessage.message = `${error.key}:${error.message}`;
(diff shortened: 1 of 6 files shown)


pankaj-ag (commented):
Looks good to me. Can you make sure you trim the keywords before sending them to the backend?

mdshamoon:
> Looks good to me. Can you make sure you trim the keywords before sending them to the backend?

Yes, They are trimmed from the frontend.

pankaj-ag:
Can you please write the test case also?
Both Unit and cypress?

mdshamoon:
> Can you please write the test case also?
> Both Unit and cypress?

Written unit test cases. @AkhileshNegi is working on the cypress test cases.

PR #993, opened 2021-02-11
Added subscriptions for collection count and bspbalance

mdshamoon:
## Summary

Added separate subscriptions for collection count and bspbalance


pankaj-ag (approved):
Looks good to me.

pankaj-ag (approved):

PR #1003, opened 2021-02-14
Search fixes

kurund:
## Summary

* Set the default contact to 25 and messages to 20 for search queries
* Remove unnecessary search call on page load
* Code optimization and clean up
* Removed count from savedsearch query. Update the count only using the subscription. (backend might add new api later this week or so)

## Test Plan

* Included in the PR


DigneshGujarathi (commented):
Looks fine

mdshamoon (approved):

PR #1032, opened 2021-02-18
fixed websocket error

mdshamoon:
## Summary

Added renew token functionality if we get WebSocket error.


DigneshGujarathi (commented):
Looks good

PR #1037, opened 2021-02-19
Added fixes for load more issue

DigneshGujarathi:


mdshamoon:
## Feedback

The offset is getting added even if I am on a different savedSearch.

mdshamoon:
Looks good.

PR #1042, opened 2021-02-20
Package upgrade

kurund:
## Summary

- [x] Upgrade packages
- [x] Code clean up
- [ ] Fix the failing test case
- [ ] Check the functionality in UI

## Test Plan

* Included in the PR


kurund:
Note for the future:

`date-fns` was upgraded to the latest 1.x  instead of 2.x as `Material UI picker` 3.x does not work with the latest.

https://github.com/mui-org/material-ui-pickers/issues/1440

We can move to `date-fns` 2.x after the stable release of `Material UI picker` 4.x

DigneshGujarathi (approved):
Looks good

PR #1051, opened 2021-02-22
Updated error validation

DigneshGujarathi:


src/components/floweditor/FlowEditor.tsx
```diff
@@ -260,12 +258,18 @@ export const FlowEditor = (props: FlowEditorProps) => {
     setFlowValidation('');
   };
 
-  const errorMsg = (
+  const errorMsg = () => (
```
mdshamoon, line 261: Is there any specific reason for this to be a function?

DigneshGujarathi: Yes, by default we are not having any value for flowValidation so to avoid on load call/assignment to this, I have created a function.

mdshamoon:
Looks good.

PR #1056, opened 2021-02-23
Loading background on transparent image fixes

mdshamoon:
## Summary

Fixed loading is shown on the background of the transparent image


src/graphql/mutations/Chat.ts
```diff
@@ -18,6 +18,7 @@ export const CREATE_AND_SEND_MESSAGE_MUTATION = gql`
           label
         }
         media {
+          id
```
DigneshGujarathi, line 21: You've added Media Id, but except for mock data, it is not used anywhere.

mdshamoon: @DigneshGujarathi I was earlier using the id for updating the image but later used ref. Removed all media Id's.

DigneshGujarathi (approved):
Looks good

PR #1061, opened 2021-02-23
Count query fail checks added

mdshamoon:
## Summary

Added fixes if the count query does return anything.


DigneshGujarathi (approved):
Looks good

PR #1082, opened 2021-02-26
Collection fixes for staff role

mdshamoon:
## Summary

Staff role should not have access to edit and delete functionality for a collection.


DigneshGujarathi (commented):
Looks good

PR #1087, opened 2021-02-26
Pino moved from dev dependencies to dependencies

mdshamoon:
## Summary

Pino moved from dev dependencies to dependencies


DigneshGujarathi (approved):
Looks good

PR #1099, opened 2021-03-01
Updated with flow editor's latest version

pooja1425:


mdshamoon:
yarn.lock file should also be checked in with the updated version, otherwise there will be an error with the mismatch in versions during the build time.

mdshamoon:
Looks good

PR #1100, opened 2021-03-02
now we auto renew subscriptions too.. finally

kurund:
## Summary

* Auto-renew token for subscriptions


DigneshGujarathi (commented):
Looks good

PR #1102, opened 2021-03-02
Image loading bug fixed

mdshamoon:
## Summary

Added react-image package to load image since it was very difficult to handle loading and error states manually


DigneshGujarathi (commented):
Looks good

PR #1103, opened 2021-03-02
Session popup

DigneshGujarathi:

### DIFF
--- a/src/containers/Auth/Logout/Logout.tsx
+++ b/src/containers/Auth/Logout/Logout.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useContext, useEffect, useState, CSSProperties } from 'react';
 import axios from 'axios';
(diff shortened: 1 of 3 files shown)


src/containers/Auth/Logout/Logout.tsx
```diff
@@ -10,6 +10,20 @@ import { USER_SESSION } from '../../../config';
...
 import { DialogBox } from '../../../components/UI/DialogBox/DialogBox';
 
+const divStyle: CSSProperties = {
+  height: '60px',
+  left: '37.03%',
+  right: '37.11%',
+  top: 'calc(50% - 60px/2 - 184px)',
+  fontFamily: 'Heebo',
+  fontStyle: 'normal',
+  fontWeight: 'normal',
```
mdshamoon, line 20: Any reason why we are not using modular css.

mdshamoon (approved):
Looks good

PR #1176, opened 2021-03-15
Updated tab icons

DigneshGujarathi:


mdshamoon:
Can we have the icons as clickable too?

DigneshGujarathi:
> Can we have the icons as clickable too?

Let's do it in a separate issue

mdshamoon (approved):

PR #1197, opened 2021-03-18
Block simulator bug resolved

mdshamoon:
## Summary

Simulator contacts can't be blocked

## Test Plan

Not needed.


DigneshGujarathi (commented):
Looks good

PR #1198, opened 2021-03-18
Bug fixed for empty message

mdshamoon:
## Summary

Empty message bug fixed

### DIFF
--- a/src/containers/Chat/ChatMessages/ChatInput/ChatInput.tsx
+++ b/src/containers/Chat/ChatMessages/ChatInput/ChatInput.tsx
@@ -100,7 +100,7 @@ export const ChatInput: React.SFC<ChatInputProps> = (props) => {
   }
 
   const submitMessage = (message: string) => {
-    if (!message) return;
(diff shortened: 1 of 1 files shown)


src/containers/Chat/ChatMessages/ChatInput/ChatInput.tsx
```diff
@@ -100,7 +100,7 @@ export const ChatInput: React.SFC<ChatInputProps> = (props) => {
   }
 
   const submitMessage = (message: string) => {
-    if (!message) return;
+    if (!message || message === ' \n') return;
```
DigneshGujarathi, line 103: If we add a space and then press enter, the message is still sent.
Logic should be, in my view, "remove space and check the length."

DigneshGujarathi (changes requested):
It's not allowing to send sticker & audio file. if an attachment is present empty message is allow.

DigneshGujarathi (approved):
Looks good

PR #1202, opened 2021-03-19
Bug fix for new subscription on status

mdshamoon:
## Summary

Bug fix for new subscription on status


DigneshGujarathi (approved):
Looks good

PR #1210, opened 2021-03-22
Added backup image

mdshamoon:
## Summary

Changed backup image in case the correct image does not load


DigneshGujarathi (approved):
Looks good

PR #1213, opened 2021-03-22
Updated search Icon & clickable tab icon

DigneshGujarathi:


mdshamoon:
Can we have the bottom borders the way they were

image

mdshamoon (approved):

PR #1214, opened 2021-03-23
Added specific error messages in case the message is not send

mdshamoon:
## Summary
Added specific error messages in case the message is not send

## Test Plan
Added


src/containers/Chat/ChatMessages/ChatMessages.tsx
```diff
@@ -98,12 +98,10 @@ export const ChatMessages: React.SFC<ChatMessagesProps> = ({ contactId, collecti
 
   // create message mutation
   const [createAndSendMessage] = useMutation(CREATE_AND_SEND_MESSAGE_MUTATION, {
-    onError: () => {
-      setNotification(
-        client,
-        'Sorry! 24 hrs window closed. Your message cannot be sent at this time. ',
-        'warning'
-      );
+    onError: (error: any) => {
```
DigneshGujarathi, line 101: Could you also verify if the backend supports error handling in the case of a collection?

mdshamoon: Verified with backend. They are sending errors in the case of collection too.I have added support for that as well.

DigneshGujarathi (approved):
Looks good

PR #1217, opened 2021-03-23
let's add radmonness for fetch

kurund:
## Summary

* Let's trigger fetch at random interval instead of fixed time

## Test Plan

* TBD


mdshamoon:
Testing feedback:

Its calling refetch as expected but due to a large number of subscriptions the refetch call is in pending state until the rest of subscriptions get finished. 
image

So we get new data after all the subscriptions are finished.

PR #1220, opened 2021-03-24
Save selected contact data in state

DigneshGujarathi:


DigneshGujarathi:
Need v1.2 as a base branch

PR #1241, opened 2021-03-30
Added logs for search

DigneshGujarathi:


src/containers/Chat/ChatMessages/ChatMessages.tsx
```diff
@@ -286,6 +287,11 @@ export const ChatMessages: React.SFC<ChatMessagesProps> = ({ contactId, collecti
       flow: 'OUTBOUND',
     };
 
+    addLogs(`send Message To Collection-${collectionId}`, {
```
mdshamoon, line 290: Can we create a `const variables` here as well.

mdshamoon:
Looks good.

Issue #1240, opened 2021-03-30
Show navigate away popup on back button click on the flow list screen


mdshamoon:
**Describe the task**
User should get a popup on the click of the back button on flow configure screen.

![image](https://user-images.githubusercontent.com/32592458/112980239-3f0ec380-9177-11eb-96d4-638784a90671.png)

PR #1251, opened 2021-04-01
Added navigation popup

mdshamoon:
## Summary

Fix for #1240


DigneshGujarathi (commented):
Actually, it saves our changes even if we click the confirm button. In that case, it may lead to misunderstanding.

PR #1264, opened 2021-04-06
fixed images issue in chat

KumudiniK:


src/containers/Chat/ChatMessages/ChatMessage/ChatMessageType/ChatMessageType.module.css
```diff
@@ -3,7 +3,7 @@
   max-width: 280px;
   min-width: 170px;
```
mdshamoon, line 4: Can we also remove the minimum width since we get extra blank space if the image is of small width

PR #1283, opened 2021-04-12
Use glific floweditor npm package

mdshamoon:
## Summary

Switching to glific floweditor npm package.
https://www.npmjs.com/package/@glific/flow-editor

Added authorization for all API calls from the frontend itself

Also fixed one deepscan issue
image


DigneshGujarathi (approved):
Looks good.
Need to check to intercept XHR requests code.

PR #1287, opened 2021-04-13
Validation fixes

shreesh1993:
1. Now form values won't reset, when attachment type is changed.
2. Changed CSS for chat-message-type, changed fixed-height to **auto**.


DigneshGujarathi (approved):
Looks good

PR #1297, opened 2021-04-15
Now shortcode won't get resetted with old hsm state

shreesh1993:


src/containers/Template/Form/Template.tsx
```diff
@@ -348,6 +346,9 @@ const Template: React.SFC<TemplateProps> = (props) => {
       helperText: warning,
       onChange: (event: any) => {
         const val = event || '';
+        if (!event) {
+          setIsUrlValid('');
```
DigneshGujarathi, line 350: We can set val to setIsUrlValid then no need to if condition

PR #1300, opened 2021-04-16
Added username and email fields in billing page

mdshamoon:
## Summary

Added username and email fields in the stripe billing page. Now users can fill in their details on the billing page itself.

<b>Internal working</b>

1. User inputs name and email that creates a customer with createCustomer mutation.
2. Once the customer is created it attaches a payment method with that customer in the next step


DigneshGujarathi (approved):
Looks good

PR #1302, opened 2021-04-16
Update floweditor with delay node changed from action to router

mdshamoon:
## Summary

Updated floweditor with delay node changed from action to router


mdshamoon:
Closing since its already covered in #1314.

PR #1318, opened 2021-04-21
Org setup

shreesh1993:


src/containers/Organization/OnboardSuccess/OnboardSuccess.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import GlificLogo from '../../../assets/images/logo/Logo.svg';
+import styles from './OnboardSuccess.module.css';
+
+export const OnboardSuccess: React.SFC = () => {
+  console.log('onboard setup success');
```
mdshamoon, line 6: Remove console.

src/components/UI/Form/Input/Input.tsx
```diff
@@ -116,7 +116,9 @@ export const Input: React.SFC<InputProps> = ({ textArea = false, disabled = fals
           <FormHelperText className={styles.DangerText}>{form.errors[field.name]}</FormHelperText>
         ) : null}
         {helperText ? (
-          <FormHelperText className={styles.HelperText}>{helperText}</FormHelperText>
+          <FormHelperText id="helper-text" className={styles.HelperText}>
```
mdshamoon, line 119: Any reason why this id is added?

shreesh1993: Since nesting of component is too deep, and styling for specific case makes more difficult. 
We had a use case for the organization onboarding form, wherein helper text size was different and alignment was right aligned. So using conditional CSS and having id on the form field makes it easier to tackle such issues.

src/containers/Organization/Registration/Registration.tsx
```diff
@@ -0,0 +1,147 @@
...
+  phone: Yup.string().required('NGO whatsapp number is required'),
+  app_name: Yup.string().required('App name is required'),
+  api_key: Yup.string()
+    .test('len', 'Invalid API Key', (val) => val?.length === 32)
+    .required('API key is required'),
+  email: Yup.string().email().required('Email is required'),
+  shortcode: Yup.string().required('NGO shortcode url is required'),
+});
+
+const initialFormValues = {
```
DigneshGujarathi, line 98: Do we need this?

src/containers/Organization/StaticOrganizationContents/StaticOrganizationContents.tsx
```diff
@@ -0,0 +1,64 @@
...
+  subtitle: string;
+  links?: any;
+  buttonText?: any;
+  handleStep?: any;
+}
+export const StaticOrganizationContents: React.SFC<StaticOrganizationContetsProps> = (props) => {
+  const { title, subtitle, links, buttonText, handleStep } = props;
+  let preRequisites;
+  if (links) {
+    preRequisites = links.map((index: any) => {
```
DigneshGujarathi, line 19: Can you name it properly - index. it's not index here. you can rename it link

src/containers/Organization/Registration/Registration.test.tsx
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+import { act, render, screen, waitFor } from '@testing-library/react';
+import UserEvent from '@testing-library/user-event';
+import { MemoryRouter, BrowserRouter } from 'react-router-dom';
+import axios from 'axios';
+
+import { Registration } from './Registration';
+import OnboardSuccess from '../OnboardSuccess/OnboardSuccess';
```
DigneshGujarathi, line 8: Need to remove this - OnboardSuccess

DigneshGujarathi (changes requested):

src/containers/Organization/Registration/Registration.tsx
shreesh1993: Yup validation fails if we don't provide initial values. Also, formik says if initial values are absent, it will consider the component uncontrolled and throws an error.
https://formik.org/docs/api/formik#initialvalues-values

mdshamoon (approved):
Looks good. Have one UI feedback. We can target it in a separate issue.

PR #1324, opened 2021-04-22
Display messages which have been replied

DigneshGujarathi:


mdshamoon:
Overall looking good. Just one minor thing, can we make the cursor as pointer when hovering over the message that has been replied.

PR #1325, opened 2021-04-22
Add authorization header for all floweditor routes

mdshamoon:
## Summary

- Added authorization header for all floweditor routes
- Token renews automatically if it expires


DigneshGujarathi (approved):
Looks good

PR #1327, opened 2021-04-23
Updated translation for menu

DigneshGujarathi:


DigneshGujarathi:
A better approach is needed.

PR #1333, opened 2021-04-25
fixed incorrect translations, and update config only to generate engl…

kurund:
…ish files


kurund:
merging to test the localization workflow

PR #1362, opened 2021-05-05
Added support for sending local attachments 

mdshamoon:
## Summary

This PR adds the following features:

- Users will have an option to select an attachment from their machine.
- The attachment will first be uploaded on GCS and we will get a public URL of the attachment.
- The public URL is then send along with the message.

## Test Plan

Added


package.json
```diff
@@ -47,6 +47,7 @@
     "@types/uuid": "^8.3.0",
     "@types/yup": "^0.29.11",
     "ajv-keywords": "^4.0.0",
+    "apollo-absinthe-upload-link": "^1.7.0",
     "apollo-link-token-refresh": "^0.3.2",
     "apollo-upload-client": "^14.1.3",
```
kurund, line 52: @mdshamoon We should remove this if not used.

kurund (changes requested):

src/types/react-tiny-link/index.d.ts
```diff
@@ -1 +1,3 @@
 declare module 'react-tiny-link';
+
+declare module 'apollo-absinthe-upload-link';
```
kurund, line 3: Each type should be defined separately.

kurund (changes requested):

package.json
mdshamoon: Removed the package

shreesh1993 (approved):

PR #1388, opened 2021-05-14
Consulting hours

shreesh1993:


src/containers/Consulting/Consulting.tsx
```diff
@@ -0,0 +1,232 @@
...
+      },
+      onChange: (val: any) => setOrganization(val),
+    },
+    {
+      component: Input,
+      name: 'participants',
+      type: 'text',
+      placeholder: t('Participants'),
+      inputProp: {
+        onChange: (event: any) => setParticipants(event.target.value),
```
mdshamoon, line 125: Do we need this listener here? Can you also check if these are needed in others as well because formik handles it internally?

src/containers/Consulting/Consulting.tsx
```diff
@@ -0,0 +1,232 @@
...
+      name: 'when',
+      placeholder: t('Select date'),
+      onChange: (val: any) => setWhen(val),
+    },
+    {
+      component: Input,
+      name: 'duration',
+      type: 'text',
+      placeholder: t('Enter time (in mins)'),
+      validate: (val: any) => {
```
mdshamoon, line 139: Validation should be done through formik.

src/containers/Consulting/Consulting.tsx
```diff
@@ -0,0 +1,232 @@
...
+        let error: any;
+        if (val % 15 !== 0) {
+          error = 'Duration should be in multiple of 15';
+        }
+        return error;
+      },
+      inputProp: {
+        onChange: (event: any) => setDuration(Number(event.target.value)),
```
mdshamoon, line 147: Do we need this?

mdshamoon, line 147: Do we need this?

src/containers/Consulting/Consulting.tsx
shreesh1993: Yes, we need listeners because we want to have both formik and local states synced. Otherwise, values for formik state reset with our local state(since local state values are not being updated) when you change the dropdown value.

src/containers/Consulting/Consulting.tsx
shreesh1993: This was custom validation so I've added validate function. But we can replace this using `Yup.number().test() `

shreesh1993: Test functions got 3 params
- test name
- validation error message
- validation function which returns true/false

image

mdshamoon (approved):

PR #1392, opened 2021-05-17
Fixed horizontal scroll issue for flows

shreesh1993:
- Added min-width for columns
- Fixed horizontal scroll issue


src/containers/Flow/FlowList/FlowList.module.css
```diff
@@ -39,9 +40,7 @@
 }
 
 .LastPublished {
```
mdshamoon, line 42: Let's remove this if we are not using it.

mdshamoon:
@shreesh1993 can you also check for other lists screen like tag, speed sends, and triggers for this resolution.

mdshamoon (approved):

Issue #1365, opened 2021-05-06
Fix warnings in test cases and research on best practices


mdshamoon:
**Describe the task**

Fix warnings in test cases and research on best practices around that.

mdshamoon:
### Please use the following approach if you find a warning/error for the below scenarios. 

**- Formik state update error( mostly where the form is being used)** 
You can use `fireEvent.change` instead of `userEvent.type`, when input element is textbox.  Check out the below screenshot.
Also, you'll have to wrap under `waitFor` so that it will update the state.
![image](https://user-images.githubusercontent.com/26791171/118608471-5bbe9380-b7d7-11eb-86c1-78bd2b21985c.png)


**- For prop types errors**
Check for prop-type before passing value. Most errors comprise of passing value as a `string` for prop-type `boolean`. 
For example Autocomplete component with dropdown, options are expected to have value as an `empty array` but was supplied `""`

These warnings are most common in our test cases. Check out the below screenshot.
![image](https://user-images.githubusercontent.com/26791171/118606289-da660180-b7d4-11eb-9674-3caadcafea59.png)

kurund:
This [PR](https://github.com/glific/glific-frontend/pull/1473) includes several fixes related to this issue.

mdshamoon:
#1589 includes several fixes related to this issue.

PR #1398, opened 2021-05-18
Test cases warning fixes

shreesh1993:
### Please use the following approach if you find a warning/error for the below scenarios. 

**- Formik state update error( mostly where the form is being used)** 
You can use `fireEvent.change` instead of `userEvent.type`, when input element is textbox.  Check out the below screenshot.
Also, you'll have to wrap under `waitFor` so that it will update the state.
image

**- For prop types errors**
Check for prop-type before passing value. Most errors comprise of passing value as a `string` for prop-type `boolean`. 
For example Autocomplete component with dropdown, options are expected to have value as an `empty array` but was supplied `""`

These warnings are most common in our test cases. Check out the below screenshot.
image


mdshamoon:
Added the description in this [issue](https://github.com/glific/glific-frontend/issues/1365#issuecomment-849568730). @shreesh1993 let's create a new PR with these and some more test warning fixes.

PR #1428, opened 2021-06-01
Unit test cases warning fixes

shreesh1993:
- Most errors are with keys, sometime we have duplicate value for key attribute hence the issue
- Also few errors are due to the state being updated and we are not waiting for that. Wrap component with act()..

### DIFF
--- a/src/containers/Chat/ChatMessages/AddVariables/AddVariables.tsx
+++ b/src/containers/Chat/ChatMessages/AddVariables/AddVariables.tsx
@@ -95,10 +95,9 @@ export const AddVariables: React.FC<AddVariablesPropTypes> = ({
             alignButtons="left"
           >
             <div data-testid="variablesDialog">
-              {formFieldItems.map((field: any, index: any) => {
-                const key = index;
-                return <Field {...field} key={key} />;
-              })}
+              {formFieldItems.map((field: any) => (
+                <Field {...field} key={field.name} />
+              ))}
             </div>
           </DialogBox>
         </Form>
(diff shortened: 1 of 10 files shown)


src/containers/Chat/ChatMessages/AddVariables/AddVariables.tsx
```diff
@@ -95,9 +95,10 @@ export const AddVariables: React.FC<AddVariablesPropTypes> = ({
             alignButtons="left"
           >
             <div data-testid="variablesDialog">
-              {formFieldItems.map((field: any) => (
-                <Field {...field} />
-              ))}
+              {formFieldItems.map((field: any, index: any) => {
+                const key = index;
```
mdshamoon, line 99: We should not use index as keys. Can you check if we have something unique like a name that can be added.

src/containers/Chat/ChatMessages/ChatMessages.tsx
```diff
@@ -582,37 +582,40 @@ export const ChatMessages: React.SFC<ChatMessagesProps> = ({ contactId, collecti
...
-          );
-          if (messageTags.length > 0) {
-            messageTags = messageTags[0].tags;
+    reverseConversation = reverseConversation.map((message: any, index: number) => {
+      const key = index;
+      return (
+        <ChatMessage
+          {...message}
+          contactId={contactId}
+          key={key}
```
mdshamoon, line 591: Same here. Message id will always be unique. Why are we using index here?

shreesh1993: The issue was mock data, which basically had the same `message.id`

mdshamoon (approved):

PR #1454, opened 2021-06-08
Page does not refresh on edit contact fields

mdshamoon:
## Summary

Page does not refresh on edit contact fields


mdshamoon:
Targeting this later since the approach does not look good for updating the cache.

PR #1456, opened 2021-06-08
Package upgrade

kurund:
## Summary

- [x] Upgrade packages
- [x] Code clean up
- [x] Fix the failing unit and cypress test cases
- [x] Check the functionality in UI

## Test Plan

* Included in the PR


kurund:
A note for the future:

`date-fns` was upgraded to the latest 1.x  instead of 2.x as `Material UI picker` 3.x does not work with the latest.

https://github.com/mui-org/material-ui-pickers/issues/1440

We can move to `date-fns` 2.x after the stable release of `Material UI picker` 4.x

src/containers/Profile/User/UserProfile.module.css
```diff
@@ -1 +1,2 @@
-.UserProfile {}
\ No newline at end of file
+.UserProfile {
```
mdshamoon, line 1: Let's remove this if not using it.

mdshamoon (approved):

PR #1459, opened 2021-06-10
Template button support

shreesh1993:
- Added support to add template buttons
- You can choose between call-to-action or quick reply buttons
- You can have max 3 quick reply buttons
- You can have max 2 call to actions buttons, but type should not be same(ie URL and phone)
- Realtime update on simulator
- Validations for phone no and url
- Now when edited buttons will show up


mdshamoon:
Please check integration test errors. Especially for speed sends.

mdshamoon (approved):

PR #1478, opened 2021-06-23
Fixed variables values getting vanished when focus is lost on chat

shreesh1993:
- Fixed variables on chat screen getting vanished once focus is lost


mdshamoon:
Testing feedback:
@shreesh1993 if we reopen the dialog after clicking on done. It does not restore the values.

mdshamoon (approved):

PR #1485, opened 2021-06-28
Now simulator will support attachments

shreesh1993:
- Added support for attachment on simulator


mdshamoon:
@shreesh1993 testing feedbacks:

- The box should close once we click on any of the options. Currently, it remains open even after sending the attachment
image

- The document should have a title. Currently, it is an empty string.

mdshamoon (approved):

PR #1490, opened 2021-06-29
UI/responsive dialogbox

KumudiniK:


src/config/theme.tsx
```diff
@@ -75,4 +75,24 @@ const theme = createMuiTheme({
   },
 });
 
+theme.typography.h6 = {
+  fontSize: '1.25rem',
+  '@media (max-width:768px)': {
+    fontSize: '1rem',
+  },
+  [theme.breakpoints.down('sm')]: {
+    fontSize: '1rem',
```
mdshamoon, line 84: Any reason why we are using both @media and theme.breakpoints. Do we need them both?

src/containers/Extensions/Extensions.module.css
```diff
@@ -5,7 +5,8 @@
 .Dialogbox {
   border-radius: 32px !important;
   margin: 0px;
-  min-width: 456px;
+  min-width: 280px;
+  max-width: 436px;
```
mdshamoon, line 9: Why did we change this? It's looking like this on the desktop screen
image
 You can use media queries in this case

mdshamoon (approved):

PR #1498, opened 2021-07-04
Code fixes

kurund:

### DIFF
--- a/src/components/UI/Form/Calendar/Calendar.tsx
+++ b/src/components/UI/Form/Calendar/Calendar.tsx
@@ -42,6 +42,7 @@ export const Calendar: React.SFC<CalendarProps> = ({
     <MuiPickersUtilsProvider utils={DateFnsUtils}>
       <Grid className={styles.Calendar}>
         <KeyboardDatePicker
+          error={hasError ? errorText : ''}
           autoOk
           variant={variant}
           inputVariant={inputVariant}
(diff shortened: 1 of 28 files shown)


src/components/UI/Form/Calendar/Calendar.tsx
```diff
@@ -42,7 +42,6 @@ export const Calendar: React.SFC<CalendarProps> = ({
     <MuiPickersUtilsProvider utils={DateFnsUtils}>
       <Grid className={styles.Calendar}>
         <KeyboardDatePicker
-          error={hasError ? errorText : ''}
```
mdshamoon, line 45: This is needed to show the border as red on an error in the Calendar component.
It can be changed to `error={hasError}`
image

src/components/UI/Form/TimePicker/TimePicker.tsx
```diff
@@ -43,7 +43,6 @@ export const TimePicker: React.SFC<TimePickerProps> = ({
       <Grid className={styles.TimePicker}>
         <KeyboardTimePicker
           autoOk
-          error={hasError ? errorText : ''}
```
mdshamoon, line 46: Same here

src/components/UI/Form/Calendar/Calendar.tsx
kurund: error is not actually the property of KeyboardDatePicker hence I had removed it. I am reverting this for now till we find a better solution.

src/components/UI/Form/TimePicker/TimePicker.tsx
kurund: Reverted this too. I think it works as it passes all props to the underlying text component.

mdshamoon (approved):

PR #1496, opened 2021-07-02
Added interactive message list and form structure

mdshamoon:
## Summary

Added interactive message list and form structure

PR #1513, opened 2021-07-07
Task/interactive msg tab

KumudiniK:


src/containers/Chat/ChatMessages/ChatInput/ChatInput.tsx
```diff
@@ -92,7 +94,8 @@ export const ChatInput: React.SFC<ChatInputProps> = (props) => {
           data.createMessageMedia.messageMedia.id,
           attachmentType,
           selectedTemplate,
-          variableParam
+          variableParam,
+          null
```
mdshamoon, line 98: If this is not passed then also it assumes the parameter to be null

src/containers/Chat/ChatMessages/ChatInput/ChatInput.tsx
```diff
@@ -160,8 +163,12 @@ export const ChatInput: React.SFC<ChatInputProps> = (props) => {
           },
         },
       });
+    } else if (type) {
+      props.onSendMessage(null, null, type, null, null, message);
+      setSelectedTab(''); // unselect the the tab
+      setOpen(false); // close the tab list suggestions
     } else {
-      props.onSendMessage(message, null, 'TEXT', selectedTemplate, variableParam);
+      props.onSendMessage(message, null, 'TEXT', selectedTemplate, variableParam, null);
```
mdshamoon, line 171: We don't need to pass null at the end.

src/containers/Chat/ChatMessages/ChatInput/ChatInput.tsx
```diff
@@ -195,14 +202,17 @@ export const ChatInput: React.SFC<ChatInputProps> = (props) => {
...
     resetVariable();
 
-    // set selected template
-    setSelectedTemplate(obj);
-
-    // Conversion from HTML text to EditorState
-    setEditorState(EditorState.createWithContent(WhatsAppToDraftEditor(obj.body)));
+    if (isInteractiveMsg) {
+      submitMessage(obj.interactiveContent, obj.type);
+    } else {
```
mdshamoon, line 210:
```suggestion
    if (isInteractiveMsg) {
      submitMessage(obj.interactiveContent, obj.type);
      return;
    } else {
```

Let's add a return in if since we don't need to run the further lines if it goes in if

src/containers/Chat/ChatMessages/ChatTemplates/ChatTemplates.tsx
```diff
@@ -8,38 +8,77 @@ import { ReactComponent as AttachmentIconUnselected } from '../../../../assets/i
...
+          case 'image':
+            tabListToShow = interactiveJSON.content.caption;
+            break;
+          case 'video':
+            tabListToShow = interactiveJSON.content.caption;
+            break;
+          case 'file':
+            tabListToShow = interactiveJSON.content.filename;
+            break;
+          default:
```
mdshamoon, line 64:
```suggestion
      } else if (interactiveJSON.type === 'quick_reply') {
       const {content} = interactiveJSON
        switch (content.type) {
          case 'text':
            tabListToShow = content.text;
            break;
          case 'image':
            tabListToShow = content.caption;
            break;
          case 'video':
            tabListToShow = content.caption;
            break;
          case 'file':
            tabListToShow = content.filename;
            break;
          default:
```

mdshamoon:
@KumudiniK I have resolved the feedbacks in #1496.
Closing this in favor of #1496

PR #1521, opened 2021-07-09
fixes for autocomplete implementation

kurund:
## Summary

* Autocomplete field generates several warnings although it's working. This PR attempts to clean up implementation and the places where it is used.

## Test Plan

* Included in the PR


src/containers/InteractiveMessage/InteractiveOptions/InteractiveOptions.tsx
```diff
@@ -62,6 +62,7 @@ export const InteractiveOptions: React.SFC<InteractiveOptionsProps> = ({
     if (templateType === LIST) {
       template = (
         <ListReplyTemplate
+          key={index}
```
shreesh1993, line 65: Is it okay to use the index as a key? or should we go for just a random string such as the current date? Sometime back when I used index as key, deepscan gave me an error :(

src/containers/StaffManagement/StaffManagment.test.tsx
```diff
@@ -30,7 +30,7 @@ test('it should have a help link', async () => {
   // loading is show initially
   expect(screen.getByText('Loading...')).toBeInTheDocument();
   await waitFor(async () => await new Promise((resolve) => setTimeout(resolve, 0)));
-  const helpButton = screen.getByTestId('helpButton');
+  const helpButton = await screen.getByTestId('helpButton');
```
shreesh1993, line 33: I've updated this test case, wrapped code under `await waitFor` in the Shamoon branch.
note: **await** is used with **findBy**

shreesh1993 (approved):

src/containers/InteractiveMessage/InteractiveOptions/InteractiveOptions.tsx
kurund: Nope, we should not be directly using the index of the map for the key. I have updated the code.

src/containers/StaffManagement/StaffManagment.test.tsx
kurund: I don't see any open PR's., what branch?

> note: await is used with findBy

Not sure what you mean here. Let's discuss tomorrow.

kurund: Fixed.

kurund:
@shreesh1993 

Merging this as per our discussion.

Issue #1529, opened 2021-07-13
Feedbacks on interactive messages


mdshamoon:
**Feedbacks**

- [x] Body/Footer is not present for quick replies
- [x] If we switch between list and quick reply it deletes the previous data

PR #1531, opened 2021-07-13
Interactive/contact field support

### LINKED ISSUE #1524: Add contact variable support for interactive messages
Currently, we don't have a dropdown that will show the contact variables list

- Add dropdown support for contact variables when user types the `@` symbol

https://www.sketch.com/s/ba323b85-11d7-4687-8a19-681a32132bab/a/JnWZrpG

shreesh1993:
- Added contact variable support when users enter @ symbol
- Now when you switch between template types, old values will be preserved

closes #1524 , #1529


src/components/UI/Form/Input/Input.tsx
```diff
@@ -120,6 +122,7 @@ export const Input: React.SFC<InputProps> = ({ textArea = false, disabled = fals
             {helperText}
           </FormHelperText>
         ) : null}
+        {dropdown && dropdown}
```
mdshamoon, line 125:
```suggestion
        {dropdown}
```

If its null then it would not show.

PR #1533, opened 2021-07-14
Added import flows functionality

mdshamoon:
## Summary

Added import flows functionality


kurund:
@mdshamoon @shreesh1993 

* We should handle errors gracefully. Export existing flow and then try to import the same. The form just breaks with bad errors.
Screenshot 2021-07-22 at 10 20 25 PM

I am not able to import a flow.

src/components/UI/Form/EmojiInput/EmojiInput.tsx
```diff
@@ -48,7 +48,7 @@ const customSuggestionsFilter = (searchValue: string, suggestions: Array<any>) =
   return filteredSuggestions.slice(0, length);
 };
 
-const DraftField = React.forwardRef((inputProps: any, ref: any) => {
+const DraftField = (inputProps: any) => {
```
kurund, line 51: I am not sure about these changes? Is this related to the import functionality?

kurund (changes requested):
Need better error handling.

src/components/UI/Form/EmojiInput/EmojiInput.tsx
shreesh1993: The focus was somehow **shifting** on  **Emoji** Input when the user is typing elsewhere in the form. And we mostly using **ref** to focus emoji input. And I did not find a use case wherein we could potentially want to focus Emoji input.

shreesh1993:
> @mdshamoon @shreesh1993
> 
> * We should handle errors gracefully. Export existing flow and then try to import the same. The form just breaks with bad errors.
>   Screenshot 2021-07-22 at 10 20 25 PM
> 
> I am not able to import a flow.

It seems you don't have **importflow** mutation, could you please take the latest master pull from the backend and check again? 
Make sure you do `mix setup` and `mix phx.server`

kurund:
Sorry, my bad. I was missing the mutation. After the backend reset, I no longer get the error on import. However, the import does not work. 

Screenshot 2021-07-23 at 9 41 47 AM

src/components/UI/Form/EmojiInput/EmojiInput.tsx
kurund: Ok

kurund:
I added the below notifications to make it more clear to end users.

Screenshot 2021-07-23 at 11 32 11 AM
Screenshot 2021-07-23 at 11 31 40 AM

PR #1586, opened 2021-08-10
Removed unnecessary logs and added import flow text

mdshamoon:
## Summary

Removed unnecessary logs and added import flow text


mdshamoon:
Self-reviewed and merged.

PR #1588, opened 2021-08-11
Added functionality for Locking flows while editing 

mdshamoon:
## Summary

Added functionality for Locking flows while editing 

## Test Plan

Added

### DIFF
--- a/src/containers/Flow/FlowList/FlowList.tsx
+++ b/src/containers/Flow/FlowList/FlowList.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useRef } from 'react';
+import React, { useState, useRef, useEffect } from 'react';
 import { Link, useHistory } from 'react-router-dom';
 import moment from 'moment';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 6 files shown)


src/containers/Flow/FlowList/FlowList.tsx
```diff
@@ -68,6 +74,12 @@ export const FlowList: React.SFC<FlowListProps> = () => {
   const [flowName, setFlowName] = useState('');
   const [importing, setImporting] = useState(false);
 
+  const [releaseFlow] = useLazyQuery(RELEASE_FLOW);
+
+  useEffect(() => {
+    releaseFlow();
```
kurund, line 80: Any specific reason why we are always calling releaseFlow() on the listing screen?

kurund (approved):
Functionality is working as expected.

kurund:
@mdshamoon 

We should wait for https://github.com/glific/glific/pull/1694 to merge and then we can merge this PR. Without that, it will break the frontend master flow.

src/containers/Flow/FlowList/FlowList.tsx
mdshamoon: We have many options on the floweditor screen that can be used to release flow namely:

1. After publishing
2. On click of the `back` button
3. On Click of `save as draft` button
4. When we close the tab

So rather than adding release flow on each of these actions, I thought this would be a much more optimal way. Let me know your thoughts.

kurund: ok, that's fine.

PR #1594, opened 2021-08-16
Added name of the person who is updating flows in the dialog message description

mdshamoon:
## Summary

Added name of the person who is updating flows in the dialog message description. 

## Test Plan

Added


kurund (approved):
Looks good and is working as expected. However, we should merge this only after https://github.com/glific/glific/pull/1704 is merged as it breaks frontend.

PR #1597, opened 2021-08-16
Fixed custom element error in floweditor

mdshamoon:
## Summary

Fixed custom element error in floweditor


kurund:
@mdshamoon 

Let's keep this PR on hold as per our discussion.

kurund (commented):
on hold for now.

kurund (approved):
The current workaround seems to be reasonable. We have not heard back from the package maintainers. It is working as expected, so good to merge.

PR #1609, opened 2021-08-19
Removed tags code in advanced search and removed unnecessary margins

mdshamoon:
## Summary

Removed tags code in advanced search and removed unnecessary margins


kurund (approved):
looks good and is working as expected!

PR #1612, opened 2021-08-20
Add more logs on logout and gracefully logout from floweditor

mdshamoon:
## Summary

Add more logs on logout and gracefully logout from flow editor


mdshamoon:
Closing this since it's not needed for now.

PR #1619, opened 2021-08-27
Fixed interactive message not able to save bug

mdshamoon:
## Summary

Fixed interactive message not able to save bug


kurund (approved):
Looks good to me!

PR #1641, opened 2021-09-10
Added lazy loading for chat components

### LINKED ISSUE #1639: Login screen shows blank page. - Performance
**Describe the bug**
Sometimes, the login screen doesn't come up. When the user accesses the Glific login URL Ex. (https://cc.tides.coloredcow.com/login), it shows a blank page and a refresh of the page immediately shows the login page. The issue is intermittent, doesn't happen all the time.

**Additional comments**
As per discussion in standup. need to check the below things from a performance perspective.
1. Check the Search queries we fire at the time of loading the page. See if there is anything that can be avoided at the time of loading.
2. Optimize the queries.

mdshamoon:
## Summary

- Added lazy loading for chat and chat subscription component since the login page loads take much time because of that
- Partially resolves #1639


kurund (approved):
@mdshamoon Looks good.

PR #1690, opened 2021-10-06
Fixed draft js formatting issues

mdshamoon:
## Summary

Added decorator for bold and removed markdown formatting
Formattting will now be similar to whatsApp

### DIFF
--- a/src/common/RichEditor.tsx
+++ b/src/common/RichEditor.tsx
@@ -1,9 +1,6 @@
 import React from 'react';
 import reactStringReplace from 'react-string-replace';
-import {
-  EditorState,
-  ContentState,
-} from 'draft-js';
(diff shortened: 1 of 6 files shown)


kurund:
type 'test'

then add * around it. text is converted to bold and * are still shown.

It is behaving correctly in simulator

mdshamoon:
> Here is an odd behavior:
> 
> Before fix: Screenshot 2021-10-07 at 10 25 47 PM
> 
> After fix: Screenshot 2021-10-07 at 10 26 20 PM
> 
> Ideally, the text should not be bold and if we are making it bold then * should not be shown

@kurund I was keeping it like how WhatsApp does it but yeah I think this can confuse people. Let me remove it.

PR #1726, opened 2021-10-29
Updated help link in flow editor screen

mdshamoon:
## Summary

Updated help link in flow editor screen.


mdshamoon:
We need to add a redirection to docs.glific.org/flow here. On hold until this is done.

mdshamoon:
> We need to add a redirection to docs.glific.org/flow here. On hold until this is done.

Hardcoding the URL since we don't have a way to make this redirection right now.

kurund (approved):
looks good

PR #1737, opened 2021-11-02
Safari mobile browser does not show the header part

mdshamoon:
## Summary

Fixed Safari browser header issue

**Note:** Calc() function of CSS was not working the same way in Safari as in Chrome so used javascript to calculate height and added inline. 

For testing use the Simulator app in MacOS.


kurund (changes requested):
@mdshamoon 

* Working fine on Chrome and also in mobile view.
* On Safari browser, contacts are not shown after clicking on Chat icon on right.
Screenshot 2021-11-29 at 11 55 48 PM

mdshamoon:
@kurund fixed the clicking issue on Safari. I used the check with window.screen.width which in Safari returns the system screen width. Replaced it with window.innerWidth

kurund (approved):
The fix is working fine, mobile Safari, desktop Safari and also mobile view on Chrome.

PR #1778, opened 2021-12-14
Added code-ql workflow

mdshamoon:
## Summary

Added code-ql workflow


mdshamoon:
This is not needed since we already have synk for vulnerability detection in the code.

PR #1827, opened 2022-01-06
Fixed timer component

mdshamoon:
## Summary

- Updated timer component to only select time by UI.
- Added validation for time in default flow option under organization settings


kurund (changes requested):
@mdshamoon 

 I am still able to enter an invalid date. Which throws backend error.

Screenshot 2022-01-15 at 12 58 16 PM

mdshamoon:
@kurund checked the material UI pickers and it uses the native browser time pickers, that does not have every browser support. So going with the keyboardDateTImePicker for now. Added validation for startTime as well.

kurund (approved):
looks good

PR #1829, opened 2022-01-06
send mail

pankaj-ag:
Enabled send_email node in the flow editor


mdshamoon:
As discussed in the standup, we don't need this right now. Closing.

PR #1836, opened 2022-01-07
Subscription will now only be present in chat and after opening simulator in preview

mdshamoon:
## Summary

Subscription will now only be present in chat and after opening simulator in preview


kurund (approved):
Looks good. Great job with clean up.  Ready to merge 🤞

PR #1839, opened 2022-01-09
cancel button fixes

kurund:


mdshamoon (changes requested):
This reduces the border for login and configure button as well which should not be changed. 
image

mdshamoon (approved):

PR #1845, opened 2022-01-12
Disable name editing in profile page

mdshamoon:
## Summary

Disable name editing in profile page


kurund (changes requested):
@mdshamoon 

Name editing is disabled on Contact Profile. However, logged in user can still change their name.

I am not sure if it's expected behaviour. Let's discuss.

PR #1858, opened 2022-01-17
menu fixes

kurund:


mdshamoon:
@kurund just wanted to check if we should close the menu when we move the cursor out.

kurund:
@mdshamoon 

I have kept the behaviour same as before. May be good to close.

mdshamoon (approved):

PR #1866, opened 2022-01-20
Added null check for contact fields

mdshamoon:
## Summary

Added null check for contact fields


mdshamoon:
Self reviewed and merged

PR #1874, opened 2022-01-24
Added all day check in organization settings

mdshamoon:
## Summary

Added all day check in organization settings


mdshamoon:
> @mdshamoon
> 
> Working as expected. However, getting the below error message in the console after saving it.
> 
> ```
> index.js:1 Missing field 'organization' while writing result {
>   "__typename": "Organization",
>   "activeLanguages": [
>     {
> ```

Cheked the issue. Seems its already there before these changes. Also checked for resolution but it seems to be an issue with the apollo client. https://github.com/apollographql/apollo-client/issues/8677. Everyhting on our end in the code seems to be fine.

kurund (approved):
Ok, let's go ahead and merge then.

PR #1891, opened 2022-02-03
Hotfix/ Removed validation

mdshamoon:
## Summary

Removed validation for interactive messages title


mdshamoon:
@kurund removing validation for now since this is needed urgently. We can add a proper regex later that supports other languages as well

kurund (approved):
looks good!

PR #1935, opened 2022-02-22
Possible fix for websocket reconnect bug

mdshamoon:
## Summary

Added warning when socket connection breaks

### DIFF
--- a/src/common/notification.ts
+++ b/src/common/notification.ts
@@ -3,11 +3,12 @@ import { NOTIFICATION, ERROR_MESSAGE } from 'graphql/queries/Notification';
 export const setNotification = (
   client: any,
   message: string | null,
-  severity: string | null = 'success'
+  severity: string | null = 'success',
+  hideDuration: number | null = 5000
 ) => {
   client.writeQuery({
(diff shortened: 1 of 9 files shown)


kurund (changes requested):
I am seeing a weird issue after these changes. Notification alert is now displayed on the Chat screen.

Screenshot 2022-02-23 at 11 47 41 PM

Here are the steps to replicate:
1. Edit a flow
2. Open preview
3. Then click on Chat
4. After few minutes you will see notification on the Chat screen.

kurund:
I feel since we resetting WebSocket connection on clicking preview on Floweditor it's resulted in this conflict.

mdshamoon:
> WebSocket

Resetting connection was causing some unwanted behaviours so removing that for now. Also the notification is system wide since the connection can break on any screen and user needs to be aware that the live messages will not come in that case.

kurund (approved):
Reload message is now shown correctly on Chat and Floweditor screens if the WebSocket connection is lost.

I made a minor correction. Good to merge once the test passes.

PR #1943, opened 2022-02-27
menu improvements

kurund:


mdshamoon:
Checked with Akhilesh and the main issue for him is still that we have to move out of one menu to hover over another as it hides the other icon. 

Let's update the menu position as well.

mdshamoon (commented):
Checked with Akhilesh and the main issue for him is still that we have to move out of one menu to hover over another as it hides the other icon. 

Let's update the menu position as well.

kurund:
@mdshamoon 

We will need to fix the cypress test after this is merged.

kurund:
Just realised the mobile view needs some fixes. So let me fix that too.

kurund:
@mdshamoon 

Ready for your review.

mdshamoon (approved):

PR #1963, opened 2022-03-14
Import hsm template functionality added

mdshamoon:
## Summary

Import hsm template functionality added.


kurund (changes requested):
The code looks fine. However, import is not working. May be due to backend changes?

kurund (approved):
@mdshamoon 

There was an issue with my CSV. Now the import is working as expected.

Good to merge once backend changes are merged.

**Note:**
Currently, if you reimport the same CSV again after making text changes etc, it is ignored. I think it might be good to discuss this during standup.

PR #1967, opened 2022-03-17
Added copy interactive functionality

mdshamoon:
## Summary

Added copy interactive functionality


kurund (changes requested):
@mdshamoon 

There seem to be a few issues with copy action.

1. Copy default interactive message, Glfic it. It updates the existing message instead of creating a new one.
2.  Following above now for each copy action title is setas 'Copy Glific'

kurund:
@mdshamoon 

Working as expected for new ones. It would be good to fix the sample set at some point.

PR #1974, opened 2022-03-22
Added same payload as gupshup for interactive messages 

mdshamoon:
## Summary

Added same payload as gupshup for interactive messages

### DIFF
--- a/src/containers/Chat/ChatMessages/ListReplyTemplate/ListReplyTemplate.tsx
+++ b/src/containers/Chat/ChatMessages/ListReplyTemplate/ListReplyTemplate.tsx
@@ -191,7 +191,7 @@ export const ListReplyTemplateDrawer: React.SFC<ListTemplate> = (props) => {
                         value={option.title}
                         name="radio-list-item"
                         size="small"
-                        checked={checkedItem && option.title === checkedItem?.payload.title}
+                        checked={checkedItem && option.title === checkedItem.payload.title}
                         color="primary"
                       />
                     </div>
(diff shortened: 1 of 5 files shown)


src/containers/Chat/ChatMessages/ListReplyTemplate/ListReplyTemplate.tsx
```diff
@@ -158,12 +160,27 @@ export const ListReplyTemplateDrawer: React.SFC<ListTemplate> = (props) => {
         <div className={styles.Options}>
           {options
             .map((option: any) => {
+              const payloadObject = {
+                payload: {
+                  type: 'list_reply',
+                  title: option.title,
+                  id: '',
```
kurund, line 167: Should we be generating unique id here?

src/containers/Chat/ChatMessages/QuickReplyTemplate/QuickReplyTemplate.tsx
```diff
@@ -35,22 +36,35 @@ export const QuickReplyTemplate: React.SFC<QuickReplyTemplateProps> = (props) =>
...
 
   const quickReplyButtons = options
-    .map((option: ButtonOption) => {
+    .map((option: ButtonOption, index: number) => {
       if (option.title) {
+        const payloadObject = {
+          payload: {
+            type: 'button_reply',
+            title: option.title,
+            id: '',
```
kurund, line 53: Should we generate unique id?

src/containers/Chat/ChatMessages/ListReplyTemplate/ListReplyTemplate.tsx
mdshamoon: The id comes empty from the gupshup end as well. Since we are making it exactly same let's not add it for now. If needed we can add it later. 

We can also discuss this in our call.

kurund: ok

PR #1976, opened 2022-03-24
Changed default sort direction for webhook logs

mdshamoon:
## Summary

Changed default sort direction for webhook logs


kurund (approved):
looks good!

PR #1979, opened 2022-03-25
Fixed bug for interactive list focus

mdshamoon:
## Summary

Fixed bug for interactive list focus


kurund (approved):
Looks good.

Let's discuss this fix during our call.

PR #1989, opened 2022-03-29
Removed actions for ngo admin

mdshamoon:
## Summary

Removed actions for ngo admin in staff management list


mdshamoon:
Since the action is removed for NGO admin the cypress tests are failing. Will update those after this gets merged

kurund (approved):
Looks good!

PR #1991, opened 2022-03-30
Added error boundary for react

mdshamoon:
## Summary

Added error boundary for react.

Steps to test:

```
Add a constant in any file
const test: any= {}

now try to nested check for that object

if(test.button.show === true)

It will show the error boundary

```


kurund (commented):
@mdshamoon 
Displaying the error as expected. However, nothing happens on clicking the Ok button. Looks at the code it is supposed to log out the user.

Is it working for you?

mdshamoon:
> if(test.button.show === true)

Yes. Its happening due to the react-scripts-issue that I told you earlier. An iframe blocks the entire screen
https://github.com/facebook/create-react-app/issues/11773
I have added a fix written in this issue. Setting react-error-overlay to v6.0.9. This works but let's upgrade reacts-scripts soon

kurund (commented):
@mdshamoon 

I don't see any changes in behaviour, maybe we can park this till the package upgrade?

kurund:
@mdshamoon 

We might need to change the implementation as we are using the latest React.

mdshamoon:
> @mdshamoon
> 
> We might need to change the implementation as we are using the latest React.

Based on the [documentation](https://reactjs.org/docs/error-boundaries.html) React doesnt have anything new for Error Boundary in the latest version. It uses the same approach of the class component. 
Please let me know if I am missing something here.

kurund (approved):
Looks good and working as expected!

PR #1992, opened 2022-03-30
Hidden wallet balance for enterprise

mdshamoon:
## Summary

Hidden wallet balance for enterprise


kurund:
running test suite again.

Issue #1824, opened 2021-09-03
HSM should be mandatory in Send Broadcast messages 


mohitsharma-22:
**Is your feature request related to a problem? Please describe.**
NGO staff members cannot be in the session window all the time to receive session messages.
If a staff member is not in the session window, the flow gets to stops and doesn't execute after send sending the next message.

**Describe the solution you'd like**
Making HSM template mandatory will avoid issues related to stoping flow execution in between and causing confusion to NGO users.

pankaj-ag:
This is not urgent. Just transferring the issue since it's related to frontend only.

mdshamoon:
We are going with hiding the session message box and adding a message in that place to select template from WhatsApp tab

Issue #1783, opened 2021-12-15
Add option to select a sticker as an attachment in floweditor


mdshamoon:
**Describe the task**
Add option to select a sticker as an attachment in floweditor.

**References**
#1707

Issue #1720, opened 2021-10-26
Add option to select type of attachment after uploading the URL


mdshamoon:
**Describe the task**
Add option to select type of attachment after uploading the URL

kurund:
@mdshamoon 

We can select type here.

![Screenshot 2022-03-26 at 9 13 43 PM](https://user-images.githubusercontent.com/151481/160257371-71065844-854d-443d-981d-5183ef4a4f70.jpg)

Am I missing something?

mdshamoon:
@kurund there's another option to upload a URL when we have GCS activated. In that when we upload the file it does not ask the type of file to be sent.

PR #2005, opened 2022-04-05
Updated floweditor with broadcast and attachment updates

mdshamoon:
## Summary

Updated floweditor with broadcast and attachment updates.
Fixes for #1783 #1720 #1824


kurund (approved):
looks good

PR #2007, opened 2022-04-05
Added responsiveness in forms

mdshamoon:
## Summary

Added some level of responsiveness in forms


mdshamoon:
Will come back to this later if needed.

PR #2015, opened 2022-04-07
Added contacts count instead of users count

mdshamoon:
## Summary

Added contacts count instead of users count


kurund (approved):
looks good

PR #2017, opened 2022-04-07
Added expression support for search

mdshamoon:
## Summary

Added expression support for search


mdshamoon:
@kurund fixing the cypress issues in the cypress repo as we have changed the way how we interact with calendar.

kurund (changes requested):
@mdshamoon 

I don't see any option for using the expression in the UI.

Screenshot 2022-04-15 at 10 50 12 PM

kurund:
@mdshamoon @abhi1203 

I wonder if we should use `Use relative dates` instead of `Use expression` label. Thoughts?

mdshamoon:
> @mdshamoon
> 
> I don't see any option for using the expression in the UI.
> 
> Screenshot 2022-04-15 at 10 50 12 PM

@kurund for now we are only going with the expression in create search screen.

kurund (approved):
looks good.

PR #2026, opened 2022-04-18
Added emoji check in list interactive messages

mdshamoon:
## Summary

Added emoji check in list interactive messages


kurund (approved):
looks good

PR #2051, opened 2022-05-06
Fixed organization provider save bug

mdshamoon:
## Summary

Fixed organization provider save bug


kurund (changes requested):
@mdshamoon 

I was able to deactivate it without any issues. However, when checking active again I got the below error message. It was saved correctly.

Screenshot 2022-05-11 at 10 10 19 PM

kurund:
@mdshamoon 

Saving of `BigQuery` seems to be an issue.  API call results in 500 error. 

GCS works fine.

PR #2053, opened 2022-05-06
Fixed formatting issues in simulator

mdshamoon:
## Summary

Fixed formatting issues in simulator.

Added a function that checks all the valid cases for whatsApp formatting and used interweave library for better url matching inside texts.


kurund (approved):
Looks good and working as expected!

PR #2058, opened 2022-05-10
Added separate attachment for interactive messages

mdshamoon:
## Summary

Added separate attachment for interactive messages


kurund (commented):
@mdshamoon 

I am able to add different media for each language currently on master. Maybe I am missing something. Let's have a chat tomorrow.

PR #2072, opened 2022-06-03
Added roles and permissions

mdshamoon:
## Summary

Added roles and permissions


src/common/utils.ts
```diff
@@ -151,4 +151,7 @@ export const numberToAbbreviation = (numberString: string) => {
   return abbreviation;
 };
 
+// need to check from backend if organization has dynamic role
+export const organizationHasDynamicRole = () => getOrganizationServices('rolesAndPermission');
```
kurund, line 155: @mdshamoon 

Storing this in local storage might have some security implications. Let's chat a bit more if you think otherwise.

mdshamoon:
Need to change cypress test cases according to new APIs. Will fix that after merging this.

PR #2094, opened 2022-08-01
Pinned flows

kurund:


kurund:
@mdshamoon 

Can you please review this? I feel one cypress test will fail and needs to be fixed once this is merged.

src/common/constants.ts
```diff
@@ -141,7 +141,7 @@ export const setColumnToBackendTerms: any = (listName: string, columnName: strin
   };
 
   const additionalItems: any = {
-    flow: { TITLE: 'name' },
+    flow: { TITLE: 'name', ' ': 'is_pinned' },
```
mdshamoon, line 144: This does not feel right somehow. We should think about fixing this sometime soon. Creating an issue for this https://github.com/glific/glific-frontend/issues/2097

src/mocks/Flow.tsx
```diff
@@ -190,8 +191,8 @@ export const filterFlowNewQuery = {
       opts: {
         limit: 50,
         offset: 0,
-        order: 'ASC',
-        orderWith: 'name',
+        order: 'DESC',
+        orderWith: ' ',
```
mdshamoon, line 195: The order_with should be is_pinned I think. Let me know if I am missing something here.

src/common/constants.ts
kurund: Yes, it is weird :) We map using what's displayed in the UI which is not the best way. This is also an issue when it comes to localising it.  So this is only a workaround. I will take a look at clean-up once done with package stuff.

src/mocks/Flow.tsx
kurund: Yes, thanks for the fix.

PR #2098, opened 2022-08-04
Fixed screen freezes while seeing revision history

mdshamoon:
## Summary

Fixed screen freezes while seeing revision history


kurund (approved):
I  was not able to replicate it. However, the fix does not have a side effect. Maybe it's specific to browser version etc.

Issue #2093, opened 2022-07-29
Login Issue: Login prompt appears thrice before letting user login


AkhileshNegi:
**Describe the bug**
When a user is working in multiple browser tabs and in one of the tabs the user logs out the issue occurs where the user is prompted thrice to log in before letting the user login.

**To Reproduce**
1. Go to the Glific Login page and log in with credentials
2. Create a duplicate of the tab
3. Now log out in the first tab
4. Move to the second tab and click on any of the buttons
5. The user is kicked out and prompted to log in again
6. Even after login it prompts to log in again 

**Screenshots**
![Flows](https://user-images.githubusercontent.com/40158831/181697505-6dc1791c-5c6e-43e4-b12b-b9c21ecc9a58.gif)

PR #2105, opened 2022-08-08
Login Issue #2093 fixed

enigmaticmahesh:
## Summary

Login Issue #2093 Fixed

### DIFF
--- a/src/config/apolloclient.ts
+++ b/src/config/apolloclient.ts
@@ -118,7 +118,7 @@ const gqlClient = (history: any) => {
   const httpLink: any = createLink({ uri: GLIFIC_API_URL });
 
   const retryIf = (error: any) => {
-    const doNotRetryCodes = [500, 400];
+    const doNotRetryCodes = [500, 400, 401];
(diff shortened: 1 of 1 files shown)


mdshamoon (commented):
In some cases the refresh token takes time and we need to retry after the authentication fails one time. Not sure if we should not retry if authentication fails once. We should do something on the login screen end that if we reach the login screen then don't open the popup.

Or maybe there should not be a need for retry after the authentication fails.
Let's discuss this tomorrow in our call.

kurund (commented):
@enigmaticmahesh 

This issue occurs because we are auto retying to renew session in case of errors.

The current fix should be fine for the reported issue. However, we should do some more testing for the cases where in we auto-renew session using re-try. 

@mdshamoon  Can you also please test scenarios mentioned above and also the flow editor part?

Let's have chat tomorrow if you need anything

enigmaticmahesh:
@mdshamoon I changed the **ttl** value and tried in different tabs. It is working fine

mdshamoon (approved):
I have also tested at my end and the changes looks good to merge.

PR #2106, opened 2022-08-08
Error handled while user trying to remove the contact

enigmaticmahesh:
## Summary

Error was not handled when user tries to remove the contact in its profile, which has been fixed in this PR. This issue was found while debugging the current issue


mdshamoon (commented):
@enigmaticmahesh adding the error handler on the delete action is fine. But the actual issue was that the user should not have the access to delete or remove the NGO main account number.

basically we need to hide the remove action on the NGO main account number. Right now its hidden only if the number is of current user,
image

kurund (commented):
@mdshamoon 
I think we checked, and other users do not have access to delete NGO's primary contact.

@enigmaticmahesh 
Maybe you should make another user account with the role of admin and try deleting the main account?

mdshamoon:
> and other users do not have access to delete NGO's primary contact.

We don't allow deleting your own contact. So yes I think we need to login with a different admin account.

enigmaticmahesh:
Restricted the user to delete the contact by hiding the Remove button in this [Commit](https://github.com/glific/glific-frontend/pull/2106/commits/a2e9c2fc420d518dd32eac9b46425b125b1d2ad5) @mdshamoon

src/containers/Profile/Profile.tsx
```diff
@@ -58,6 +59,7 @@ export const Profile: React.SFC<ProfileProps> = ({
 
   const loggedInUserContactId = data.currentUser.user.contact.id;
   const currentUserPhone = data.currentUser.user.phone;
+  const organizationPhone = data?.currentUser?.user?.organization?.contact?.phone;
```
mdshamoon, line 62:
```suggestion
  const { user } = data.currentUser;
  const loggedInUserContactId = user.contact.id;
  const currentUserPhone = user.phone;
  const organizationPhone = user?.organization?.contact?.phone;
```

enigmaticmahesh: @mdshamoon implemented your suggestion

mdshamoon (approved):
@enigmaticmahesh changes looks good. Please fix the deepscan issue and merge.

Issue #2096, opened 2022-08-02
Add an ability to pin flows


kurund:
Currently, flows that are used in the organization settings are automatically pinned for easy access.

We should expose also allow pinning for other flows

Here is the mockup: 
In the Edit screen "Is pinned?" https://www.sketch.com/s/ba323b85-11d7-4687-8a19-681a32132bab/a/8EOgRR

Note that we should disable this checkbox for the flows that are used in organization settings.

PR #2109, opened 2022-08-10
Added Pinned option and its disability condition while editing Flow

enigmaticmahesh:
## Summary

1. Added an option to pin flow while editing the flow
2. Disable this option if the flow is pinned through organization settings.
3. Issue #2096


enigmaticmahesh:
@mdshamoon I saw some merge conflicts in this PR, I think both the code should reside and it would resolve it.

kurund (changes requested):
Can you please resolve the conflicts?

enigmaticmahesh:
@kurund resolved the conflicts

mdshamoon (changes requested):
@enigmaticmahesh can you please check the failing unit tests.

enigmaticmahesh:
@mdshamoon fixed the unit test cases

src/containers/Flow/Flow.test.tsx
```diff
@@ -53,13 +60,22 @@ it('should support keywords in a separate language', async () => {
 });
 
 it('should not allow special characters in keywords', async () => {
-  const { container, getByText } = render(flow({ params: {} }));
+  const { container, getByText, findByTestId } = render(flow({ params: {} }));
 
-  await waitFor(() => {});
-  fireEvent.change(container.querySelector('input[name="name"]'), {
+  const nameInput = await (await findByTestId('formLayout')).querySelector('input[name="name"]');
```
kurund, line 65: Can we check a better way to achieve this? The code looks bit funky :)

kurund (changes requested):

src/containers/Flow/Flow.test.tsx
enigmaticmahesh: @kurund I referred it directly from docs, so I used it. Ok, let me check for other ways.

kurund: @enigmaticmahesh I was referring to await before await which you have removed now. So it seems fine.

PR #2115, opened 2022-08-16
Fixed staff management role issues

mdshamoon:
## Summary

Fixed staff management role issues


kurund (approved):
One cannot remove the role from the UI as it's required.  However, I was able to replicate the behaviour mentioned in the issue if the role is removed from the DB.

The fixes seem fine and work as expected.

kurund:
@mdshamoon 

I have noticed the below error when you check `Can chat with contacts from assigned collection only`.

Screenshot 2022-08-17 at 23 50 15

mdshamoon:
> @mdshamoon
> 
> I have noticed the below error when you check `Can chat with contacts from assigned collection only`.
> 
> Screenshot 2022-08-17 at 23 50 15

@kurund will add this to the backend issues. Since we are getting a 500 error

PR #2116, opened 2022-08-16
Added common active item function for lists

mdshamoon:
## Summary

Added common active item function for lists


kurund (changes requested):
@mdshamoon 

- Can we move active and inactive classes at row level
- Currently, there is an issue with the listing wherein there is no active flag. For eg: Speed Sends listing, the `Inactive` class is applied.  It should default be active if the `is_active` field is missing.

kurund (approved):
Looks good.

PR #2121, opened 2022-08-17
Switching to graphql-ws for websockets

mdshamoon:
## Summary

Switching to graphql-ws for websockets. Provides automatic reconnection and stable websocket over graphql ws protocol

Backend branch for testing: https://github.com/glific/glific/pull/2389


kurund (approved):
@mdshamoon 

Looks good to me.

However, noticed a slow loading time for the staff user.

We should also test if it scales.

mdshamoon:
@kurund tested with the staff user. It works fine for me. 
Also I tested with 2500 contacts with command
`mix run priv/repo/seeds_scale.exs --contacts 2500`
 and it is working fine at this scale.

Issue #2117, opened 2022-08-17
Webhook Logs: Not showing first webhook log popup


AkhileshNegi:
**Describe the bug**
In webhook logs screen the popup in the first log is not appearing correctly while for others it is working fine

**To Reproduce**
1. Go to Flows -> View Webhook Logs
2. Click on first log's response json
3. The popup does not appear correctly

**Screenshots**

<img width="1672" alt="image" src="https://user-images.githubusercontent.com/40158831/185048753-b8f2349a-062a-403f-a76b-2410c64317bf.png">

PR #2122, opened 2022-08-18
UI fixed

enigmaticmahesh:
## Summary

Fixed Issue #2117

### DIFF
--- a/src/components/UI/Menu/Menu.tsx
+++ b/src/components/UI/Menu/Menu.tsx
@@ -70,6 +70,7 @@ const Menu: React.SFC<MenuProps> = ({
         transition
         disablePortal={placement === 'top'}
         placement={placement}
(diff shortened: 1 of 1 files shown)


src/components/UI/Menu/Menu.tsx
```diff
@@ -70,6 +70,7 @@ const Menu: React.SFC<MenuProps> = ({
         transition
         disablePortal={placement === 'top'}
         placement={placement}
+        style={{ zIndex: 1 }}
```
mdshamoon, line 73: Let's update in theme.tsx only to apply global style for this.

enigmaticmahesh: There is no global class(e.g., Mui-xx) for popper as it is not a part of material ui rather it is a 3rd party library used by react material ui @mdshamoon

Issue #2120, opened 2022-08-17
Make start a flow field searchable


mdshamoon:
**Describe the task**
I am not able to search my flow in start a flow dialog box. It is getting difficult while scrolling thorugh lots of flows
<img width="495" alt="image" src="https://user-images.githubusercontent.com/32592458/185106530-d2079813-9d17-4046-bdfd-cd1b50e1441b.png">

**Expected behavior**
The select field should be searchable like add to collection dialog

PR #2124, opened 2022-08-18
Flow search implemented

enigmaticmahesh:
## Summary

PR for Issue #2120


mdshamoon (commented):
@enigmaticmahesh can you check if we can use SearchDialogBox component here.

mdshamoon (commented):
@enigmaticmahesh placeholder `Select flow` and the bottom text is missing. If its taking too many changes for the searchDialogbox to accomodate these lets go with the previous approach only. with separate dialog box and autocomplete.
image

Also I think we will need to fix cypress test cases as well after making the changes. Since the implementation is changed a bit

enigmaticmahesh:
@mdshamoon I have fixed the cypress test for search flow changes and raised a [PR](https://github.com/glific/cypress-testing/pull/158). Please do have a look at it

mdshamoon (approved):

PR #2133, opened 2022-08-24
Fixed flow name undefined and removal of extra isActive column 

mdshamoon:
## Summary

Fixed flow name undefined and removal of extra isActive column


kurund (approved):
Looks good.

PR #2137, opened 2022-08-29
Lets not make a global change for checkbox background

mdshamoon:
## Summary

Lets not make a global change for checkbox background as it is affecting other areas. 

We can extend the checkbox on organization registration page to use in other places where white background is needed.


kurund (approved):
Looks good!

PR #2142, opened 2022-09-02
Added sync button for hsm templates

mdshamoon:
## Summary

Added sync button for hsm templates


kurund (approved):
The code looks fine and seems to be doing the right thing. However, I can't test if actual HSM  templates are fetched. I assuming it will :)

PR #2161, opened 2022-09-27
Contact management: Bulk update contacts

mdshamoon:
## Summary

Contact management: Bulk update contacts


kurund:
@mdshamoon 

Minor fixes:
- add view instructions link
- `Select .csv` sounds bit technical. Can we change it to `Select file`? Thoughts?
- Button text can we change to `icon + text` as we should use that format elsewhere?

kurund:
We should also display appropriate error messages. For eg, for new contacts backend returns below error message.

```
{"data":{"importContacts":null},"errors":[{"details":[{"error":"New contacts were found in this file. Sorry those could not be added"},{"error":"New contacts were found in this file. Sorry those could not be added"}],"locations":[{"column":3,"line":2}],"message":"All contacts could not be opted in due to some errors","path":["importContacts"]}]}
```

However, the 
Screenshot 2022-10-02 at 08 40 54
frontend error display is only.

mdshamoon:
@kurund 

1. Added view instructions link.
2. Since we have already mentioned to select a csv file, `Select file` would make much more sense. Updated it.
3. Yes I think we should generalize the icon + text button component as well. Creating a separate issue for that.
4. The error messages are not received in the Apollo error object.  I need to check with the backend team to update to the correct format for error messages.

PR #2162, opened 2022-09-29
Saved user search history on moving back to list screen

mdshamoon:
## Summary

Saved user search history on moving back to list screen.


kurund (approved):
@mdshamoon 

I have tested on multiple listings: Templates, Flows

The search term is correctly added to the URL and navigating to `edit` or `configure` and hitting the back browser button retains the search filter.

Issue #2164, opened 2022-09-30
Fix dialog box UI issue


mdshamoon:
**Describe the task**
The error description should be center aligned and have a fixed width
<img width="493" alt="Screenshot 2022-09-30 at 1 27 53 PM" src="https://user-images.githubusercontent.com/32592458/193221320-15bb105c-222b-4f5e-a222-2d7ca9267ce8.png">

saminarp:
Hello @mdshamoon  I would like to work on this issue as part of `hacktoberfest`, I was able to run it locally using the scripts in `package.json` and finally `serve -s build/`. I would like to know how can I reproduce the screen you have shared above.

mdshamoon:
@saminarp This issue will need a key to replicate. I can DM you the key on discord if interested. We also have some good first issues that you can pick. Please join our [discord](https://discord.gg/qqhVSrc5) channel if you want to contribute more and we can discuss further.

Sharing the steps to replicate on local:

1. Login with default credentials
2. Visit https://glific.test:3000/settings/bigquery

<img width="1183" alt="image" src="https://user-images.githubusercontent.com/32592458/194247368-1f185826-52a4-489d-83cf-8faebdfcb1ff.png">

3. Add the key with insufficient permission
4. See error

PR #2194, opened 2022-10-17
Dialogbox error description centered

saminarp:
## Summary

As discussed in meeting and in the following issue https://github.com/glific/glific-frontend/issues/2164, I have centred the error title `An Error Has Occured!`. The fixes has been implemented into the file `src/components/UI/DialogBox/DialogBox.module.css` 

## Test Plan

I have extended the error message to be able to set the CSS accurately. The current behaviour can be seen here

_Error with long message_
CleanShot 2022-10-17 at 02 32 45@2x
_Error with short message_
CleanShot 2022-10-17 at 02 34 06@2x

### DIFF
--- a/src/components/UI/DialogBox/DialogBox.module.css
+++ b/src/components/UI/DialogBox/DialogBox.module.css
@@ -40,6 +40,9 @@
   color: #073f24;
   padding-top: 24px !important;
   padding-bottom: 8px !important;
+  justify-content: center !important;
+  display: flex !important;
+  flex-direction: column !important;
 }
 
 .DialogTitle h2 {
@@ -49,13 +52,13 @@
 .DialogTitleCenter {
   composes: DialogTitle;
   text-align: center;
-  max-width: 395px;
+  max-width: 100%;
 }
 
 .DialogTitleLeft {
   composes: DialogTitle;
   text-align: left;
-  max-width: 395px;
+  max-width: 100%;
 }
 
 .DialogContentCenter {


src/components/UI/DialogBox/DialogBox.module.css
```diff
@@ -49,13 +52,13 @@
 .DialogTitleCenter {
   composes: DialogTitle;
   text-align: center;
-  max-width: 395px;
+  max-width: 100%;
```
mdshamoon, line 55: Let's not change the max width. We have assigned a maximum width of the dialog box title and it should not be greater than 395px.

src/components/UI/DialogBox/DialogBox.module.css
```diff
@@ -40,6 +40,9 @@
   color: #073f24;
   padding-top: 24px !important;
   padding-bottom: 8px !important;
+  justify-content: center !important;
+  display: flex !important;
+  flex-direction: column !important;
```
mdshamoon, line 45: There is a property in Dialog box that you can pass to make the title center aligned.

mdshamoon (changes requested):
We also need to set a max width of 395px to the dialog box content so that it doesn't stretch out too much.

src/components/UI/DialogBox/DialogBox.module.css
saminarp: @mdshamoon  could you elaborate on which property? I see the CSS written for `DialogBox` as follows:
```
.Dialogbox {
  max-width: 80%;
  min-width: 459px;
  margin: 0 auto;
  height: unset !important;
  max-height: 90% !important;
}
```
is it the `margin: 0 auto`

src/components/UI/DialogBox/DialogBox.module.css
saminarp: When we use `px` we are adjusting the width exactly in the size `395px` meaning that it is static. But % allows readjusting depending on the browser and content.  Should I be using the equivalent of 395px to the `%` ?

mdshamoon: No. max-width property doesn't mean it will be set to this width. It means that the width can be extended to only 395px and not beyond that, and that is the behaviour we want.

mdshamoon:
Closing since there has been no activity for a while. @saminarp let me know if you still want to work on this issue. We can reopen the PR after that

PR #2199, opened 2022-10-18
Added sheet list and form screens

mdshamoon:
## Summary

- Added sheet list and form screens
- Added a new node in floweditor that reads the sheet data
- Added Sheet management along with other flow actions
image


kurund (changes requested):
@mdshamoon 

I am getting the below error while saving the Sheet.  Note that the sheet is saved correctly.
Screenshot 2022-10-27 at 23 41 29

I get same error in console when I click on the sync icon.

kurund (approved):
Changes look good and are working as expected sheet URL.

Getting it working in flows is a bit tricky. Let's discuss this.

PR #2247, opened 2022-11-24
Fixed bug for logout when reloading after session time

mdshamoon:
## Summary

Fixed bug for logout when reloading after session time.

Steps to replicate:

- Go to backend file: api_auth_plug.ex line no 42 and change the ttl time from 30 minutes to 1
- Login to Glific and wait for one minute.
- Now reload the page and you would be logged out.

We should not be logged out if we refresh the page.


src/App.tsx
```diff
@@ -20,7 +20,7 @@ const App = () => {
   const [drawerOpen, setDrawerOpen] = useState(window.innerWidth > 768);
 
   useEffect(() => {
-    setAuthenticated(checkAuthStatusService());
+    setAuthenticated(getAuthSession('access_token') != null);
```
kurund, line 23: So with this change, will the user session ever expire?

kurund:
@mdshamoon 

If the user is idle and their session is expired then I feel refresh or any page click should log them out. Let's discuss this during stand-up.

PR #2253, opened 2022-11-29
Interactive message image fixes

kurund:
fixes: https://github.com/glific/glific-frontend/issues/2227


mdshamoon (commented):
@kurund I think the validation is already present from the backend its just that we don't wait for the validation to complete and allow saving the interactive message before that.

I don't think we need to explicitly check for webp format here.

kurund:
@mdshamoon 

I don't think backend validation is working for Interactive messages as we are able to save Stickers as mentioned in the UI.

However, in addition to the backend, it would be good to validate the same on the frontend.

One to chat during our call.

mdshamoon:
@kurund just one more thing on this if we can wait for the backend validation before saving. I can save a video with a image URL as well if I do it quickly.

kurund:
@mdshamoon 

I have updated the PR as per our discussion to disable the save action while validation is in progress.  Can you please check?

mdshamoon (approved):

PR #2254, opened 2022-11-30
Updated registration page

mdshamoon:
## Summary

Updated registration page


mdshamoon:
Need to fix the cypress test cases for registration. Will take it separately once this gets approved

kurund (changes requested):
@mdshamoon 

UI changes look good. However, I am not able to register without pre-registering.

Am I missing something?

mdshamoon:
> @mdshamoon
> 
> UI changes look good. However, I am not able to register without pre-registering.
> 
> Am I missing something?

In this case an OTP template is send which can be send if you are out of session and given consent. On our local since we don't have the template approved with gupshup we can't test it as such.

mdshamoon:
@kurund the flow is working fine. 
- To test you need to change the otp template name to commin_otp and make it as approved in database. 
- Next click on register with whatsapp. It will try to send a otp template but will not be received on whatsapp. 
- Just open the glific account as admin in a new tab and check the otp message that has been sent.
- Paste the otp in the registration tab and it will show the contact your admin screen

kurund:
@mdshamoon 

Following the above instructions, I was able to register successfully. Few issues:

- Name was not set for the contact
- `Admin` role was assigned to the registered user.

mdshamoon:
> Following the above instructions, I was able to register successfully. Few issues:
> 
> * Name was not set for the contact
> * `Admin` role was assigned to the registered user.

@kurund 

- Name of the contact is set when the user messages from their end. As for the user the name is updated in database.
- I checked with pankaj on this and he said that only the first user is set as Admin. I also tested it on my end and it is working fine for the second user. 
image

mdshamoon:
@kurund I am merging this since the backend is already deployed and the frontend changes looks good. If there will be any issues from the backend side. Will create issues separately.

kurund:
@mdshamoon 

All good! Thanks.

PR #2267, opened 2022-12-06
Added interactive message edit in floweditor

mdshamoon:
## Summary

Added interactive message edit in floweditor


kurund (approved):
Looks good

PR #2293, opened 2023-01-05
Added recaptcha v3

mdshamoon:
## Summary

Added recaptcha v3
backend PR for testing: https://github.com/glific/glific/pull/2639

Steps to test
1. Create recaptcha v3 keys from https://www.google.com/recaptcha/admin/
2. Put client key in the .env file in frontend and the secret key in .env.dev file in the backend


src/routes/UnauthenticatedRoute/UnauthenticatedRoute.tsx
```diff
@@ -15,16 +17,18 @@ export const UnauthenticatedRoute = () => {
...
-        <Route index element={<Navigate to="/logout/user" replace />} />
-        <Route path="/login" element={<Login />} />
-        <Route path="/registration" element={<Registration />} />
-        <Route path="/confirmotp" element={<ConfirmOTP />} />
-        <Route path="/resetpassword-phone" element={<ResetPasswordPhone />} />
-        <Route path="/resetpassword-confirmotp" element={<ResetPasswordConfirmOTP />} />
-        <Route path="/organization-registration" element={<OrganizationRegistration />} />
-        <Route path="/*" element={<Navigate to="/logout/user" replace />} />
-      </Routes>
+      <GoogleReCaptchaProvider reCaptchaKey={RECAPTCHA_CLIENT_KEY}>
```
kurund, line 20: @mdshamoon 

We have added Recaptcha for all non-authenticated routes but not validating the same.

For eg: Recaptcha is shown on login and forgot password but there is no validation for the same.

Secondly, I think we need to handle the scenario when Recaptcha expires on the page. I think backed should provide appropriate error message and frontend can display the same (like refresh page)

kurund (changes requested):
It is working fine for registration and also validating the same.

It would be good to fix the comments as mentioned earlier.

src/routes/UnauthenticatedRoute/UnauthenticatedRoute.tsx
mdshamoon: Based on our discussion I have updated the PR with the error message from the backend and keeping the catcha script for all non authenticated routes.

PR #2313, opened 2023-02-03
Vite migration

kurund:
- [x] Migrate to Vite
- [x] Fix dev mode
- [x] Fix cypress test building
- [x] Unit test fixes
- [x] CI fixes
- [x] Build process fixes
- [ ] eslint fixes
- [x] Deepscan fixes
- [x] Errors when we put logflare key in env
- [x] CSS compose issue


tsconfig.json
```diff
@@ -1,34 +1,28 @@
...
+      "@testing-library/jest-dom",
+      "vite-plugin-svgr/client"
     ],
-    "typeRoots": [
-      "./src/types",
-      "./node_modules/@types"
-    ],
-    "allowJs": true,
+    "typeRoots": ["./src/types"],
+    "target": "es2022",
```
mdshamoon, line 10: Should we have target es5 as the support for es2022 will be less I think

mdshamoon (commented):
Not sure why but lesser integration tests are running in this branch. Checking the cypress branch

tsconfig.json
kurund: We should target `ESNext `

kurund: I have updated the config.

mdshamoon:
@kurund added all the environment variables and deployed on staging. Now getting this error when I open chat screen.
image
I have added stream polyfill but somehow still not working. Checking other solutions for this.

mdshamoon (approved):
LGTM 👍.

PR #2319, opened 2023-02-10
Fix sheet row count and UI for chats

mdshamoon:
## Summary

- Sheet row count  will be updated in the UI once we click on sync sheet
- Updated height on chat screen because of an issue when we open dev console


kurund (approved):
Looks good

PR #2322, opened 2023-02-13
Fixed conversations issue

mdshamoon:
## Summary

The root cause for the issue:

onCompleted is getting called on every render which appends the same contacts at the end of the list after we click on loadmore. If contacts with same id's are appended to the list the apollo cache invalidates and restores the original data hence causing this issue.

**Things I tried:**

Using useLazyQuery's promise approach but it was not working and there are a few issues with that in apollo client. I have used the alternative that was suggested in the comments on the apollo client issues i.e. to use the client object directly.
With this it seems to be working fine.


kurund (changes requested):
@mdshamoon 

The issue still persists. Here are the steps to replicate.

1. Go to chat and click on the second contact.
2. Click contact load more
3. Click on any of newly added contact's in the list.
4. Contact clicked and additional 1 contact after above contact is duplicated.

mdshamoon:
@kurund there are two issues that are causing this: 
1. onCompleted is called on every click after we have clicked on Loadmore that adds duplicate contacts.
2. The search query is also returning some duplicate contacts with a different offset (This seems to happen only on local and not on production)

I have fixed the first case in this branch and created issue for the second in the backend https://github.com/glific/glific/issues/2696.
Let me know if we can merge this and tackle that separately

kurund (approved):
looks good

PR #2326, opened 2023-02-14
Fixed issue with formatting when we have HTML tag brackets <>

mdshamoon:
## Summary

Fixed issue with formatting when we have HTML tag brackets <>


mdshamoon:
This also removes the formatting for bold and italics in WhatsApp. For now not going with this fix

PR #2328, opened 2023-02-15
Added interactive message footer

mdshamoon:
## Summary

Added interactive message footer


kurund (changes requested):
@mdshamoon 

Overall looks good. Just one minor issue. As we type in the footer the preview is not updated.

mdshamoon:
Yes @kurund we update the changes with onBlur event for title body and footer. onChange was causing some issues as to how we have implemented the adding of form items. Creating a separate issue to track this.

PR #2332, opened 2023-02-16
Removed react as a dependency in all files

mdshamoon:
## Summary

After React 17 we don't need to include React as a dependency in a file. Cleaning up from all files
Also removed ColorPicker component since we are no more using it.


kurund (approved):
Looks good

PR #2348, opened 2023-02-27
Added error message if error occurs while uploading

mdshamoon:
## Summary

Added error message if error occurs while uploading


kurund (approved):
looks good

PR #2350, opened 2023-02-28
Improve platform responsiveness for smaller screens

mdshamoon:
## Summary

Improve platform responsiveness for smaller screens


src/containers/List/List.tsx
```diff
@@ -245,7 +245,7 @@ export const List = ({
     filterItemsQuery,
     {
       variables: filterPayload(),
-      fetchPolicy: 'cache-and-network',
```
kurund, line 248: Not sure how this is related to responsiveness?

src/containers/List/List.tsx
```diff
@@ -259,7 +259,7 @@ export const List = ({
 
   useEffect(() => {
     refetchCount();
-  }, [filterPayload, searchVal, filters]);
```
kurund, line 262: Same as above

kurund (changes requested):

src/containers/List/List.tsx
mdshamoon: When we are opening or closing the sideDrawer this query is getting called again if we have this fetch-policy. I have checked multiple scenarios like adding deleting or editing a list item and there are no issues caused by changing this policy.

src/containers/List/List.tsx
mdshamoon: refetchcount was getting called even if the filterPayload and filter were not changing while closing or opening the side drawer. Removing these doesnt seem to cause any issue with the existing features.

kurund (approved):
Looks good and working as expected.

PR #2353, opened 2023-03-02
Added stricter typescript support

mdshamoon:
## Summary

Added stricter typescript support


mdshamoon:
Based on our discussion we will try to add types on the files we work on moving forward

PR #2356, opened 2023-03-07
Fixed styling issues with chatinput

mdshamoon:
## Summary

Fixed issues with chatinput


kurund (approved):
Looks good besides one minor issue which could be tackled in a separate issue.

On smaller devices when Voice is sent, the preview occupies the whole screen hence there is no way to send the message.

Screenshot 2023-03-09 at 08 20 21

PR #2359, opened 2023-03-09
Fix: Contact list in message are not in chronological order 

mdshamoon:
## Summary

Fix: Contact list in message are not in chronological order


mdshamoon:
@kurund I am not able to replicate. Let's discuss this on a call

PR #2361, opened 2023-03-15
Removed ununsed dependency that imports webpack

mdshamoon:
## Summary

Removed ununsed dependency that imports webpack


kurund (approved):
Looks good.

PR #2368, opened 2023-03-27
package upgrades

kurund:


mdshamoon:
@kurund I have fixed all the issues and test cases. Some unhandled exceptions are coming up in the github CI for unit test which is not replicating in my local. Can you please check on your system if the same issue is present.

I have also updated cypress for the fix in the test cases here https://github.com/glific/cypress-testing/pull/164

mdshamoon:
@kurund there is one remaining issue in this PR that I will tackle separately. For now I have ignored that error.

mdshamoon (approved):

PR #2375, opened 2023-04-02
Collection contact filter based on name or phone

pankaj-ag:
Target issue: https://github.com/glific/glific-frontend/issues/2376

1. Changed the filter parameter to **term** from **name** in the contact collection list screen's search. It filters the contacts based on name or phone.


mdshamoon (approved):
LGTM 👍

PR #2418, opened 2023-06-12
Docker

dlobo:
## Summary

Dockerizing Glific

## Test Plan

Get a few #c4gt students to test it :)


package.json
```diff
@@ -68,7 +68,7 @@
   },
   "scripts": {
     "setup": "yarn install --force && yarn run floweditor",
-    "dev": "vite --host=glific.test",
+    "dev": "vite --host=0.0.0.0",
```
mdshamoon, line 71: Can we create separate command to run for docker with this host.

vite.config.ts
```diff
@@ -56,8 +56,8 @@ export default ({ command, mode }: ConfigEnv): UserConfigExport => {
         open: true,
         port: 3000,
         https: {
-          key: fs.readFileSync('../glific/priv/cert/glific.test+1-key.pem'),
-          cert: fs.readFileSync('../glific/priv/cert/glific.test+1.pem'),
+          key: fs.readFileSync('cert/glific.test+1-key.pem'),
```
mdshamoon, line 59: We need to modify Readme as well to make sure people add the certificates in the frontend directory as well

mdshamoon (changes requested):

PR #2431, opened 2023-07-04
added dropdown in create template screen for selecting tag

nishant25062002:
I have added a dropdown for selecting tags on create template page.
dropdown_in_template


mdshamoon:
Closign in favor of #2449

PR #2541, opened 2023-09-10
do not evaluate html elements

kurund:
we should format only elements supported by WhatsApp


kurund:
@mdshamoon 

Integration test failures are not related to this issue. However, we should fix them separately.

PR #2608, opened 2023-10-31
Added last login fix for gupshup enterprise

mdshamoon:
## Summary

Added last login fix for gupshup enterprise


kurund (approved):
Looks good!

Issue #2420, opened 2023-06-09
Templates: Make a copy


AkhileshNegi:
### Description
Right now, when a template is rejected NGO partner needs to apply for the template again by copy-pasting content from the rejected template

### Goals
Add an option to `Make a Copy` of the template which will basically copy all the content from the existing template and show it in edit mode. The content then can be updated before submitting the template for approval

### Project Name
Glific

###  Organization Name
Project Tech4Dev

### Domain
Social Sector

### Complexity
Easy

### Tech Skills Needed
Typescript, React

### Name of Mentors
@mdshamoon 

### Category
Feature

### Product Set-Up
https://github.com/glific/

### Acceptance Criteria
PR merged into the main branch

manu31shukla:
This feature enhancement caught my attention, and I believe my skills in Typescript and React align perfectly with the requirements for this task. With prior experience in similar projects, I am confident in my ability to add the proposed option, allowing NGO partners to copy rejected templates and edit them before resubmission.  I am already familiar with the environment and setup of the projects similar to Glific project, which will enable me to dive into the task promptly and deliver efficient results. Please assign me this issue.

ZosTera2438:
glific
glific-frontend
@mdshamoon can you assign this issue to me

mdshamoon:
Hey @ZosTera2438 can you setup Glific backend and frontend on your local first and let me know. I can then assign you this issue. Also please join our discord channel for easier communication https://discord.gg/rcBxTA4wnf

jayesh9747:
@mdshamoon  hello ,  sir . My name is jayesh and i have a good knowledge of typescript and database and how to make API. can you assign me this issue ??

Singh-Supreet:
@mdshamoon  I am very confident that I will be able to do this efficiently. I have a lot of experience in making responsive websites. Looking forward to discussing with you

ishaaaa8:
Glific
complexity : easy
@mdshamoon , I'm interested in contributing to this issue. Could you please grant me permission to work on it? Thanks!

AKSHATkumar22:
Hello Sir, my name is Akshat Kumar as I have worked and built many projects as a fresher and learner wanted to work on this issue and wanted to resolve the issue. As, I think with my tech skills knowledge I have can add some value to this project and complete the project. So, Sir please let me give a chance to showcase and perform under your guidance.

mdshamoon:
[...]

PR #2618, opened 2023-11-14
Templates copy

kurund:
Fixes for #2420


kurund:
@mdshamoon 

Unit tests are working fine on my local. It's failing due to an error specific to CI which you mentioned during our call.

Cypress error is fixed in https://github.com/glific/cypress-testing/pull/167

We would also need to update the CI on the frontend after we merge this and Cypress PR. I can do that later.

mdshamoon (changes requested):
Overall code looks good.
When I copy a template I am not able to edit it. Ideally after copy I should be able to edit it.

kurund:
> When I copy a template I am not able to edit it. Ideally after copy I should be able to edit it.

This should be fixed.

PR #2676, opened 2023-12-29
add missing mutation to update organization default language

kurund:


mdshamoon (approved):
Ohh it was missing for a while and nobody reported. Thanks for checking this.

PR #2683, opened 2024-01-03
FIX: Partial Loading of Different Screen before the Login Alert Page upon Token Expiration

### LINKED ISSUE #2671: Login: UI Fixes
**Describe the bug**
After a new build or when the old login token expires the login screen fluctuates, partially loading a different screen first before going back to the login page

**To Reproduce**
1. Go to any organization URL where login tokens have expired or after a new build
2. The issue can be seen in the Recording added

**Expected behavior**
Check if the tokens are correct and show a banner that `You have been logged out and relogin` or something sort of message in the main login page rather than partially loading different screen first

**Screenshots**
https://github.com/glific/glific-frontend/assets/40158831/13489d0e-2ee3-4a4a-a7d2-54c482f2648f

Snehil-Shah:
Fixes #2671 

## Summary

Upon login token expiration, the app now gracefully loads the Login expired alert, instead of partially loading a different page

## Test Plan

I actually had some problem in reproducing the login token expiration behaviour in development. 
So I took the approach of corrupting the Session's cookie from the devtools by changing the expiration date & renewal token ids to imitate token expiration.
This might be the reason the "Before" recording looks different (and worse) than the recording in the original issue.
I think it should work the same for actual token expiration in production as it relies on the function `checkAuthStatusService` which primarily checks for token expiration and it handles it pretty well.

I have attached **video demos** of this issue before & after this PR:

[Before](https://drive.google.com/file/d/11UiSXlaKR2xYFjB-cXCydjsD0S_rY6FS/view?usp=sharing) (watch till the end)
[After](https://drive.google.com/file/d/1KPVfh6UVuDR0H0IxjPyHMHODOx2ER8Pq/view?usp=sharing)


Snehil-Shah:
[Updated Video Preview](https://drive.google.com/file/d/1t-nm3tkfcYZxWMBQULliXspiHbBCw30l/view?usp=sharing)

src/App.tsx
```diff
@@ -21,9 +38,22 @@ const App = () => {
...
-    setAuthenticated(isTokenExpired || isAccessTokenPresent);
+    const checkAuth = async () => {
+      const isAccessTokenPresent = getAuthSession('accessToken') !== null;
+      let isSessionAlive = false;
+      if(!isAccessTokenPresent){
+        // New user
+        isSessionAlive = false;
+      }
+      else{
+        // existing user
```
mdshamoon, line 49: We should also check if the token time is expired or not. Only do a renew call if the token time is expired.

src/App.tsx
```diff
@@ -21,9 +38,22 @@ const App = () => {
...
+        isSessionAlive = await checkSessionValidity();
+      }
+      startTransition(() => {
```
mdshamoon, line 52: Do we need startTransition? I don't see how this will be helpful in our context.

mdshamoon (changes requested):

src/App.tsx
Snehil-Shah: The tests were raising warnings and recommending the use of StartTransition.

src/App.tsx
Snehil-Shah: @mdshamoon is it okay now?

src/App.tsx
```diff
@@ -21,9 +38,27 @@ const App = () => {
...
 
   useEffect(() => {
-    const isAccessTokenPresent = getAuthSession('accessToken') !== null;
-    const isTokenExpired = checkAuthStatusService();
-    setAuthenticated(isTokenExpired || isAccessTokenPresent);
+    const checkAuth = async () => {
+      const isAccessTokenPresent = getAuthSession('accessToken') !== null;
+      const isTokenAlive = checkAuthStatusService();
+      let isSessionAlive = false;
+      if(!isAccessTokenPresent){
```
kurund, line 45: Code formatting is incorrect. @Snehil-Shah Can you please check your code editor configuration?

src/App.tsx
```diff
@@ -21,9 +38,27 @@ const App = () => {
...
+        // New user
+        isSessionAlive = false;
+      }
+      else if(isTokenAlive){
```
kurund, line 49: same as above

src/App.tsx
```diff
@@ -21,9 +38,27 @@ const App = () => {
...
+        // Healthy Session
+        isSessionAlive = true;
+      }
+      else{
```
kurund, line 53: same as above

kurund (changes requested):

src/App.test.tsx
```diff
@@ -62,4 +62,86 @@ describe('<App /> ', () => {
...
+      expect(renewAuthToken).toHaveBeenCalled();
+    });
+
+    await waitFor(() => {
+      expect(screen.getByTestId('navbar')).toBeInTheDocument();
+    });
+  });
+
+  // it('it should render <Login /> component and try to renew token if session has expired', async () => {
+  //   vi.mock('services/AuthService', async (importOriginal) => {
```
mdshamoon, line 113: Remove commented test case

src/App.tsx
```diff
@@ -7,12 +7,34 @@ import 'assets/fonts/fonts.css';
...
+  renewAuthToken,
+  setAuthSession,
+  checkAuthStatusService,
+} from 'services/AuthService';
 import { UnauthenticatedRoute } from 'routes/UnauthenticatedRoute/UnauthenticatedRoute';
 import { AuthenticatedRoute } from 'routes/AuthenticatedRoute/AuthenticatedRoute';
 import { Logout } from 'containers/Auth/Logout/Logout';
 import { isGreaterThanLgBreakpoint } from 'common/utils';
 
+const checkSessionValidity = async function () {
```
mdshamoon, line 22: Move this to utils.ts

mdshamoon (commented):
Please resolve merge conflicts and feedbacks. Otherwise looks good. Will merge after the changes are done

src/App.test.tsx
Snehil-Shah: @mdshamoon I actually needed some help with that, this test and the test above it, are both correct. but I think their mocks are conflicting.
when I comment one of them out, the other one works (so the tests are definitely correct), but both of them don't work together, I tried to `resetAllMocks` beforeEach, but still am not able to make it work

mdshamoon:
@Snehil-Shah closing this in favor of #2829. Integration tests were not running against a forked branch. So created a separate branch with these changes. Overall looks good. Merging it after the tests run successfully.

FYI for the test case: vi.mock does not work specific to a test. It updates the module for the whole file. So if you used it twice it picks up the latest one for both the test cases. A better way is to use vi.spy which I have added in the PR.

Issue #2684, opened 2024-01-03
FIx warning while running dev server


mdshamoon:
**Describe the task**
FIx warning while running dev server:

<img width="1017" alt="image" src="https://github.com/glific/glific-frontend/assets/32592458/7f5178e5-773c-49d2-b48f-d863fcc43b68">

PR #2692, opened 2024-01-04
warning fixes during dev mode

kurund:
Fixes: https://github.com/glific/glific-frontend/issues/2684


mdshamoon (changes requested):
@kurund I think this is affecting the cypress pipeline. Can you please check that as well

mdshamoon (approved):

Issue #2612, opened 2023-11-03
Emails: Setting low balance threshold


AkhileshNegi:
**Is your feature request related to a problem? Please describe.**
We have added support in the backend to set low balance threshold i.e. the balance when NGO receive the email of wallet balance being low in their Gupshup account. This can differ from organization to organization depending on the scale and frequency of messaging.

**Describe the solution you'd like**
Add UI for updating low balance threshold for an organization

mdshamoon:
@AkhileshNegi do we have an API for this.

PR #2697, opened 2024-01-11
organisations settings

### LINKED ISSUE #2613: Emails: Add toggle for warning mails
**Is your feature request related to a problem? Please describe.**
We have added support in the backend to toggle on/off warning mails i.e. In addition to critical notification, an organization will get email for warnings as well

**Describe the solution you'd like**
Add UI for toggling this on/off

akanshaaa19:
## Summary
Added the following: 
- Toggle for warning mails
- Low balancethreshold
- critical balance threshold
closes #2613 #2612 #2703 

image


mdshamoon (changes requested):
Overall looks good. Few changes that I think we should do-

- Send warning email checkbox should be the first item
- If checkbox is unchecked we should disable the input fields for low and critical balance threshold. (You can refer to flows section under settings)

src/containers/SettingList/Organisation/Organisation.test.tsx
```diff
@@ -48,14 +67,49 @@ test('it renders component in edit mode', async () => {
   // loading is show initially
   expect(getByText('Loading...')).toBeInTheDocument();
 
+  //correct values are rendered in the form
   await waitFor(() => {
-    const phoneNumber = getByTestId('phoneNumber');
+    const inputElements = screen.getAllByRole('textbox');
+    const numberInputElements = screen.getAllByRole('spinbutton');
+
+    const orgName = inputElements[0] as unknown as HTMLInputElement;
```
mdshamoon, line 75:
```suggestion
    const orgName = inputElements[0] as HTMLInputElement;
```
Any reason we are converting it to unknown first?

src/containers/SettingList/Organisation/Organisation.test.tsx
```diff
@@ -48,14 +67,49 @@ test('it renders component in edit mode', async () => {
...
+    fireEvent.change(inputElements[0], { target: { value: 'Glific' } });
+    fireEvent.change(inputElements[1], { target: { value: 'Please change me, NOW!' } });
+    fireEvent.change(numberInputElements[0], { target: { value: '10' } });
+    fireEvent.change(numberInputElements[1], { target: { value: '5' } });
+  });
+
+  await waitFor(() => {
+    const submitButton = screen.getByText('Save');
+    expect(submitButton).toBeInTheDocument();
+    fireEvent.click(submitButton);
```
mdshamoon, line 113: Events should always be outside of waitFor
and the end of a test should always be an assertion

mdshamoon (commented):
Overall looks good. Added some comments, please check.

src/containers/SettingList/Organisation/Organisation.test.tsx
akanshaaa19: because it is an HTMLElement so it doesn't have value prop

mdshamoon: But we can directly do `as HTMLInputElement`. Why do we need to add `as unknown`

mdshamoon (approved):

PR #2733, opened 2024-02-02
fixed send location issue

akanshaaa19:

### DIFF
--- a/src/components/simulator/Simulator.tsx
+++ b/src/components/simulator/Simulator.tsx
@@ -167,7 +167,7 @@ export const Simulator = ({
     let context: any = {};
 
     if (interactivePayload) {
-      type = interactivePayload.payload.type;
+      type = interactivePayload.payload.type || interactivePayload.type;
       payload = interactivePayload.payload;
       delete payload.type;
       context = interactivePayload.context;
--- a/src/components/simulator/Simulator.tsx
+++ b/src/components/simulator/Simulator.tsx
@@ -167,7 +167,7 @@ export const Simulator = ({
     let context: any = {};
 
     if (interactivePayload) {
-      type = interactivePayload.payload.type || interactivePayload.type;
+      type = interactivePayload.payload.type;
       payload = interactivePayload.payload;
       delete payload.type;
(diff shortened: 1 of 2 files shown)


src/components/simulator/Simulator.tsx
```diff
@@ -167,7 +167,7 @@ export const Simulator = ({
     let context: any = {};
 
     if (interactivePayload) {
-      type = interactivePayload.payload.type;
+      type = interactivePayload.payload.type || interactivePayload.type;
```
mdshamoon, line 170: Can we use sendMediaMessage function to send the location request message similar to how we are doing while using the location button in attachment list. 

Refer implementation of the location button in the attachment list menu inside simulator.

mdshamoon (changes requested):

src/components/simulator/Simulator.tsx
akanshaaa19: but wouldn't there be inconsistency if we do this, because `sendMediaMessage` this function is used for all the attachment messages and this `sendMessage` for the other kind of messages (like interactive messages)

mdshamoon: Let's discuss this once in tomorrow's call. I think its an overlap of the two types.

src/containers/Chat/ChatMessages/ChatMessage/LocationRequestTemplate/LocationRequestTemplate.tsx
```diff
@@ -29,6 +20,7 @@ export const LocationRequestTemplate = ({
   onSendLocationClick = () => {},
 }: LocationRequestTemplateProps) => {
   const body = content.body.text;
+  const locationPayload = SAMPLE_MEDIA_FOR_SIMULATOR[5];
```
mdshamoon, line 23: We just need to send the payload part of SAMPLE_MEDIA. I tried this in a flow and its not working as expected.

mdshamoon (changes requested):

mdshamoon (approved):

PR #2736, opened 2024-02-02
Quick reply limit

### LINKED ISSUE #2721: Template: Limit check for quick replies
**Describe the task**
While creating a template message and adding quick replies, the submissions are being rejected due to quick replies exceeding 25 characters. To streamline the process and prevent template rejections for this reason, please implement a validation in the template creation form, ensuring that quick reply options adhere to the 25-character limit. This will help save time by minimizing unnecessary back-and-forth.
here is the [issue](https://discord.com/channels/717975833226248303/1201824049425022986/1201824049425022986)
**steps to replicate**
1. Login to Glific
2. go to template and create a template
3. check the add button option and add a text with more than 25 characters

**Expected Behaviour**
There should be a limit check for quick replies

akanshaaa19:
## Summary
close #2721

### DIFF
--- a/src/containers/TemplateOptions/TemplateOptions.tsx
+++ b/src/containers/TemplateOptions/TemplateOptions.tsx
@@ -235,7 +235,9 @@ export const TemplateOptions = ({
               touched.templateButtons &&
               touched.templateButtons[index] ? (
                 <FormHelperText>{errors.templateButtons[index]?.value}</FormHelperText>
-              ) : null}
+              ) : (
+                <FormHelperText>{'Quick replies limited to 25 characters.'}</FormHelperText>
+              )}
             </FormControl>
           </div>
           <div>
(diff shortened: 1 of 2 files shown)


src/containers/TemplateOptions/TemplateOptions.tsx
```diff
@@ -235,7 +235,9 @@ export const TemplateOptions = ({
               touched.templateButtons &&
               touched.templateButtons[index] ? (
                 <FormHelperText>{errors.templateButtons[index]?.value}</FormHelperText>
-              ) : null}
+              ) : (
+                <FormHelperText>{'Quick replies limited to 25 characters.'}</FormHelperText>
+              )}
```
mdshamoon, line 240: Why do we need this? The error message is already added in the validation

mdshamoon (approved):

PR #2750, opened 2024-02-12
Testing cypress fixes

mdshamoon:
## Summary

Testing cypress fixes


mdshamoon:
@kurund I remembered why we moved cypress out in the CI pipeline. 
It was because of the PR comment that comes for cypress. It takes the PR context only when we are in the same folder which is checked out. If we move out of that folder it doesnt add the PR comment. I tried multiple ways to do this if we keep cypress separately but couldnt find a good way. Will check back on this later for now I am adding it back since the PR comment helps in checking the issues easily

PR #2790, opened 2024-03-06
Start a flow in whatsapp groups

akanshaaa19:


src/containers/Chat/ChatMessages/ContactBar/ContactBar.tsx
mdshamoon, line 1: At this point I think we should change the name of this component and keep this generic. We should move out all contact related data to a parent component and resuse the same for group.

mdshamoon (changes requested):

mdshamoon (approved):

PR #2794, opened 2024-03-08
added triggers for whatsapp groups

akanshaaa19:


src/containers/Trigger/TriggerType/TriggerType.tsx
```diff
@@ -0,0 +1,60 @@
...
+          row
+          value={groupType}
+          onChange={(event) => {
+            handleOnChange(event.target.value);
+            setFieldValue(event.target.value);
+          }}
+        >
+          <div className={styles.RadioLabelWrapper}>
+            <FormControlLabel
+              value={'WABA'}
```
mdshamoon, line 38: Create constants for Waba and Wa in constants file and use everywhere

src/containers/Trigger/Trigger.tsx
```diff
@@ -378,11 +385,18 @@ export const Trigger = () => {
...
+      handleOnChange: (value: any) => setGroupType(value),
+      groupType: groupType,
+      isWhatsAppGroupEnabled: isWhatsAppGroupEnabled,
+    },
     {
       component: AutoComplete,
       name: 'groupIds',
       label: t('Select collection'),
-      options: collections.groups,
+      options: collections?.groups,
```
mdshamoon, line 399: Collection options should  be filtered based on the type of Collection. RIght now it will load all the collections.

mdshamoon (changes requested):
Collections should be filtered

mdshamoon (approved):

PR #2806, opened 2024-03-12
whatsapp group collections

akanshaaa19:


mdshamoon (approved):
Overall looks good. I think we should rename some components like AddContactsToCollection as they are contact specific and we have used in context of groups.

PR #2826, opened 2024-03-20
Fixes in whatsapp groups ui

akanshaaa19:


src/containers/WaGroups/GroupDetails.tsx/GroupDetails.tsx
```diff
@@ -4,19 +4,33 @@ import { useTranslation } from 'react-i18next';
...
 
 export const GroupDetails = () => {
   const params = useParams();
   const { t } = useTranslation();
 
   const [showDeleteDialog, setShowDeleteDialog] = useState(false);
   const [deleteVariables, setDeleteVariables] = useState<any>();
+  const [groupDetails, setGroupDetails] = useState<any>();
+
+  const { loading: groupDataLoading } = useQuery(GET_WA_GROUPS, {
```
mdshamoon, line 27: We should use Get_Wa_group it should have a filter for group id

src/containers/WaGroups/GroupDetails.tsx/GroupDetails.tsx
```diff
@@ -137,13 +151,17 @@ export const GroupDetails = () => {
     );
   }
 
+  if (groupDataLoading) {
+    return <CircularProgress data-testid="loading" />;
```
mdshamoon, line 155: I think we have a loading component please reuse that in all places if possible.  We have configured that to be center aligned on the screen. This adds a loading at the top left of the screen which looks weird

mdshamoon (changes requested):

mdshamoon (approved):

PR #2830, opened 2024-03-26
Fetch more conversations

akanshaaa19:
Things in this PR
- load more messages were duplicating the messages
- on switching from `/chat -> /group/chat `or `/group/chat -> /chat` messages were duplicating
- added fetchMore for fetching more messages or contacts instead of calling the query again


src/common/utils.ts
```diff
@@ -191,7 +191,12 @@ export const getDisplayNameForSearch = (conversation: any, groups?: boolean) =>
   if (groups) {
     displayName = conversation.name || conversation.waGroup?.label || '';
   } else {
-    displayName = conversation.name || conversation.maskedPhone || conversation.contact?.name || '';
+    displayName =
+      conversation?.name ||
+      conversation?.maskedPhone ||
+      conversation?.contact?.maskedPhone ||
+      conversation?.contact?.name ||
```
mdshamoon, line 198: Let's pass the contact object only inside this function rather than checking for maskedPhone and name again inside contact.

mdshamoon (commented):
Looks fine overall. Can you fix the codebeat issues. Specifically the duplication and the block nesting too deep once. You can ignore the others

src/mocks/Collection.tsx
```diff
@@ -345,28 +345,28 @@ export const getCollectionContactsQuery = {
   },
 };
 
-export const getCollectionInfo = {
+// export const getCollectionInfo = {
```
mdshamoon, line 348: Remove if not needed

mdshamoon (approved):
Have tested it out for most scenarios. Seems to be working fine. Just remove this commented code.

PR #2836, opened 2024-03-29
Enhancements in WA Groups

akanshaaa19:
- Changed the api for getting whatsapp group label
- Added notification for errors in sending messages


src/containers/WaGroups/GroupDetails.tsx/GroupDetails.tsx
```diff
@@ -24,24 +24,12 @@ export const GroupDetails = () => {
   const [deleteVariables, setDeleteVariables] = useState<any>();
   const [groupDetails, setGroupDetails] = useState<any>();
 
-  // we should use wa_group api with filter by id here
-  const { loading: groupDataLoading } = useQuery(GROUP_SEARCH_QUERY, {
+  const { loading: groupDataLoading } = useQuery(GET_WA_GROUP, {
```
mdshamoon, line 27:
```suggestion
  const { data: groupDetails, loading: groupDataLoading } = useQuery(GET_WA_GROUP, {
```

Use data object rather than creating another state variable

mdshamoon (changes requested):

src/containers/Collection/CollectionList/CollectionList.tsx
```diff
@@ -300,7 +297,9 @@ export const CollectionList = () => {
             navigate(`/${groups ? 'group/' : ''}collection/add`);
           },
         }}
-        filters={filter}
+        filters={{
+          groupType: groups ? WA_GROUPS_COLLECTION : CONTACTS_COLLECTION,
+        }}
```
mdshamoon, line 302: Let's not change this and keep the filters as a state variable. 
filters is used as a dependency array in one of the useEffect inside List component that calls the fetch and count API. If you define it this way then on each re-render it creates a new object with this group attribute which in turn triggers the useEffect and calls the list API again.

You can check this while adding a new group to a collection. If you search for a waGroup it calls the list and count api for groups again and again

mdshamoon (commented):
Minor change. Rest looks good. 
Please update and merge

kurund:
@mdshamoon 

As you already reviewed this PR, can please do the needful.

src/containers/Collection/CollectionList/CollectionList.tsx
```diff
@@ -83,6 +82,10 @@ export const CollectionList = () => {
     : COLLECTION_SEARCH_QUERY_VARIABLES;
   const updateMutation = groups ? UPDATE_COLLECTION_WA_GROUP : UPDATE_COLLECTION_CONTACTS;
 
+  useEffect(() => {
+    setFilter(groups ? WA_GROUPS_COLLECTION : CONTACTS_COLLECTION);
```
mdshamoon, line 86:
```suggestion
    setFilter({groupType: groups ? WA_GROUPS_COLLECTION : CONTACTS_COLLECTION});
```

When you add groupType inside the filter prop it creates an new object on every render and calls the filter and count query again.

mdshamoon (changes requested):

mdshamoon (approved):

PR #2864, opened 2024-04-18
Fixed the buttons in interactive messages

### LINKED ISSUE #2863: Interactive message buttons
Screenshots
image

akanshaaa19:
resolves #2863


mdshamoon (commented):
I think its not related to interactive messages but all messages in general. Its happening in case the flow name is too big. Check the message with text hello, it should not have so much width.
image

PR #2888, opened 2024-05-12
Changed file paths for code coverage

akanshaaa19:


kurund (commented):
As discussed over the call. Let's keep this on hold for now and explore an alternate approach.

PR #2905, opened 2024-05-30
Updated floweditor version

akanshaaa19:


kurund:
@akanshaaa19 

Do fix Deepscan issue after discussing the patch with @mdshamoon

src/components/floweditor/FlowEditor.helper.ts
```diff
@@ -106,7 +108,7 @@ export const setConfig = (uuid: any) => {
 };
 
 export const loadfiles = (startFlowEditor: any) => {
-  (window as any).static_url = '/sitestatic/';
+  (window as any).static_url = '/static/';
```
kurund, line 111: Can you check if we need this? If not then just remove.

src/components/floweditor/FlowEditor.tsx
```diff
@@ -206,11 +206,6 @@ export const FlowEditor = () => {
       Track('Flow opened');
 
       return () => {
-        Object.keys(files).forEach((node: any) => {
```
kurund, line 209: @mdshamoon 

It would be good to get your thoughts on this.

mdshamoon: When we are loading the floweditor it adds all the js and css files from the build folder to the document body. On unmounting the component, we are just removing those css and js files. the files object holds the node information.

If you inspect the document you will see these files appended at the end of the body.

If we remove this logic then everytime it adds the same set of files to the body without removing the previous ones.

This could only fail in the case when the node is not present in the body. I think we can add a check for that.

package.json
```diff
@@ -17,11 +17,12 @@
     "@emoji-mart/react": "^1.1.1",
     "@emotion/react": "^11.11.4",
     "@emotion/styled": "^11.11.5",
-    "@glific/flow-editor": "^1.19.3-8",
+    "@glific/flow-editor": "^1.26.3-1",
     "@lexical/react": "^0.13.1",
     "@mui/icons-material": "^5.15.15",
     "@mui/material": "^5.15.15",
     "@mui/x-date-pickers": "^7.3.1",
+    "@nyaruka/temba-components": "^0.89.0",
```
mdshamoon, line 25: I don't think this would work.

Even if you add it here it will only work for the dev server and not the production build, as it is dynamically loading from inside node_modules which will not be present when we do an actual build and host.
You can maybe verify by deploying this branch on staging.

Can you check how Rapidpro loads it and lets follow the same logic.

mdshamoon (changes requested):

src/components/floweditor/FlowEditor.tsx
akanshaaa19: fixed this

src/components/floweditor/FlowEditor.helper.ts
akanshaaa19: we dont, removed this

package.json
akanshaaa19: so rapidpro uses v1.19,0, which we were using earlier so we cant do that,
i tried to deploy this on staging and it worked as expected you can test once here https://staging.glific.com/flow

mdshamoon (changes requested):
Few issues that I noticed on staging

1. The name of the sheet is not appearing on opening a google sheet node
image

2. There is a lot of gap between the input box and the label. Check the styles that are added in index.html in floweditor and apply the same here
image

3. In wait for response the time appears twice after clicking on no response check 
image

PR #2908, opened 2024-05-31
Collection management

akanshaaa19:


kurund (changes requested):
* Can you please fix the alignment for the select all checkbox, it should be in line with the listing checkboxes.
Screenshot 2024-06-05 at 23 27 25

* Contact is still listed after deletion on the contact Collection listing

* On Group collection listing can we replace the add group icon instead of using the add contact icon?
Screenshot 2024-06-05 at 23 37 05

akanshaaa19:
> * Can you please fix the alignment for the select all checkbox, it should be in line with the listing checkboxes.
> 
> Screenshot 2024-06-05 at 23 27 25
> * Contact is still listed after deletion on the contact Collection listing
> * On Group collection listing can we replace the add group icon instead of using the add contact icon?
> 
> Screenshot 2024-06-05 at 23 37 05

fixed these @kurund

src/components/UI/SearchDialogBox/SearchDialogBox.tsx
```diff
@@ -68,6 +86,14 @@ export const SearchDialogBox = (props: SearchDialogBoxProps) => {
 
   const changeValue = (event: any, value: any) => {
     setSelectedOptions(value);
+    if (multiple) {
```
kurund, line 89: Looks like we are duplicating the code here (line 66)

kurund (changes requested):
Screenshot 2024-06-09 at 21 50 20

Cross allows to clear

Which results in the below behaviour.

Screenshot 2024-06-09 at 21 50 36

Then clicking add does not do anything.

I think this is bit confusing. Let's discuss during our check in.

kurund:
The below interface is not scalable. A group might have many contacts. 

Screenshot 2024-06-09 at 22 02 30

Let's discuss

src/components/UI/SearchDialogBox/SearchDialogBox.tsx
akanshaaa19: the useEffect is for when options are loaded for the first time and the function is for when option is selected

kurund (approved):
Looks good.

PR #2917, opened 2024-06-10
Fixed failing cypress tests

akanshaaa19:


akanshaaa19:
Not needed since the branch is merged now.

PR #2922, opened 2024-06-11
Enhancements in Interactive messages

akanshaaa19:


kurund (approved):
@akanshaaa19 

I have made a few code changes, and added translation support and layout fixes.

Please check and if all is good then go ahead with the merge.

kurund:
@akanshaaa19 

One outstanding issue is in edit mode `Allow dynamic media` is not checked and the user will have to check again. This does not break any functionality we can fix that after deployment.

kurund:
> @akanshaaa19
> 
> One outstanding issue is in edit mode `Allow dynamic media` is not checked and the user will have to check again. This does not break any functionality we can fix that after deployment.

I have fixed this issue. Please check. Good to merge.

PR #2927, opened 2024-06-12
Fixed validation of url in interactive messages

akanshaaa19:


kurund (approved):
@akanshaaa19 

I found an issue with my fix and have changed the logic for the dynamic media checkbox as it was always checked for valid URLs. 

Now I am checking for the valid URL instead of actually validating when it's loaded.  This should be fine as it was already validated while saving and will be validated again if changed.

PR #2936, opened 2024-06-18
Fixed message in speed sends

akanshaaa19:


kurund (changes requested):
@akanshaaa19 

Speed send form is not editable and is_active is also unchecked.

Screenshot 2024-06-19 at 22 30 03

mdshamoon (approved):

mdshamoon:
@akanshaaa19 translations for speed sends are not working. Can you please create a separate issue for that

PR #2975, opened 2024-07-11
Feature: Template flows

akanshaaa19:


src/containers/Form/FormLayout.tsx
```diff
@@ -578,6 +585,12 @@ export const FormLayout = ({
                 variant="outlined"
                 color="primary"
                 onClick={() => {
+                  if (additionalAction?.state === 'template') {
```
mdshamoon, line 588: If possible can we make a generic way to update this and not hardcode template inside formlayout component

src/containers/Flow/FlowList/FlowList.test.tsx
```diff
@@ -21,15 +22,17 @@ import { Flow } from '../Flow';
...
 const mocks = [
-  getFlowCountQuery,
-  getFlowCountQuery,
-  getFlowCountQuery,
-  getFlowCountQuery,
-  filterFlowQuery,
-  filterFlowQuery,
-  filterFlowQuery,
-  filterFlowQuery,
+  getFlowCountQuery(isActiveFilter),
```
mdshamoon, line 28: At this point can you check why there are multiple calls being made to filter and count query and see if you can reduce them. If it seems to take more time, do it in a separate PR

mdshamoon (changes requested):

akanshaaa19:
> 1. Ideally view it should directly open the flow. What would opening the flow in edit mode achieve.
> 
> Screenshot 2024-07-15 at 4 39 40 PM
> 2. Lets call the menu item templates since we are already inside flows.
> 
> Screenshot 2024-07-15 at 4 38 33 PM

on view it we will show description about the template flow thats why it takes to edit page first

mdshamoon (commented):
Please check failing unit and integration test cases. Otherwise looks good

mdshamoon (approved):

PR #2990, opened 2024-07-22
Bug: Session templates

akanshaaa19:


kurund (commented):
@akanshaaa19 

I am not able to replicate this issue on master. May be we can have a quick chat.

kurund:
@akanshaaa19 

I am able to replicate the bug now, let's not ship these changes. Let's fix the root cause.

PR #2991, opened 2024-07-22
Fix: Added not rated for templates which have unknown quality

akanshaaa19:
if a template is not rated it returns UNKNOWN, so replaced that with Not Rated

Screenshot 2024-07-22 at 7 39 48 PM


kurund (approved):
looks good to me.

PR #2992, opened 2024-07-23
Contact import automation

akanshaaa19:
Depends on backend PR https://github.com/glific/glific/pull/3666
branch: `importContact`


kurund (changes requested):
@akanshaaa19 

Overall looks good. Few suggestions:

* On clicking upload we should display the error message below Checkbox: `Are these contacts opted in?` Currently, it is shown if the user check and uncheck the Checkbox.
* We don't need to show additional notification as we have the message in helptext.
* On the below screen close does not work. It should return to `/contact-management`
Screenshot 2024-08-27 at 23 10 56
* Can we add the similar notification like about for the `Move Contact`.

I have also done minor clean up

kurund (approved):
fixes looks good!

PR #3019, opened 2024-08-12
Fixed contact read status

### LINKED ISSUE #3015: Chats are not marked as Read
**Describe the bug**
Nayi Disha has raised this query that chats are not marked as Read

**To Reproduce**
1. Go to 'Flows' >> Support Tickets 
2. Click on Send message
3. Scroll down to to type a message and send to the particular contact 
4. After sending the message, the green dot next to the name remains visible, and the unread count does not decrease.

**Expected behavior**
Once the the message is send the green dot over the contact should automatically convert to read and the unread number should also convert to read.
 
**Screenshots**
image
image
image

akanshaaa19:
fixes #3015


kurund (approved):
looks good.

PR #3030, opened 2024-08-19
Removed keyword for template flows

akanshaaa19:
target issue is https://github.com/glific/glific/issues/3702


kurund (approved):
looks good!

PR #3049, opened 2024-08-30
Fixed the input payload for updating organization flows

### LINKED ISSUE #3048: Invalid input error on updating organization even if default flow is unchecked
**Describe the bug**
Screenshot 2024-08-20 at 9 14 49 AM

The default flow is unchecked but when we try to save the changes, we are getting time invalidation error from backend

**To Reproduce**
1. Set the `out_of_office` field in organizations to ```{"enabled": false, "flow_id": null, "end_time": null, "start_time": null, "enabled_days": [{"id": 1, "enabled": false}, {"id": 2, "enabled": false}, {"id": 3, "enabled": false}, {"id": 4, "enabled": false}, {"id": 5, "enabled": false}, {"id": 6, "enabled": false}, {"id": 7, "enabled": false}], "default_flow_id": null}```
2. Go to https://glific.test:3000/settings/organization-flows
3. Click on save

**Expected behavior**
The error is coming because backend expects the start_time and end_time to be of type `time`, but FE is sending `invalid date`.
So ideally the validation is done from backend side. Maybe we can add it from FE. If out_of_office is not checked then we shouldn't send the data for updation.

akanshaaa19:
fixes #3048


kurund (approved):
Looks good.

Issue #3051, opened 2024-09-02
HSM Template: Not able to send media templates from chat screen


akanshaaa19:
**Describe the bug**
Currently we are not able to send a media template from chat screen which doesn't have a media object.
 
**To Reproduce**
1. Go to Chat screen
2. Click on Templates
3. Search for any media template and send to someone who is not in the session windo
4. See error "24 hours window closed"

PR #3052, opened 2024-09-02
Fix for media hsm templates

akanshaaa19:
target issues is #3051


src/containers/Chat/ChatMessages/ChatInput/ChatInput.tsx
```diff
@@ -182,6 +182,17 @@ export const ChatInput = ({
     // check for an empty message or message with just spaces
     if ((!message || /^\s*$/.test(message)) && !attachmentAdded) return;
 
+    //check if it has media if the template type is image
+    if (
+      selectedTemplate &&
+      selectedTemplate.isHsm &&
+      selectedTemplate.type === 'IMAGE' &&
```
kurund, line 189: Is there a reason why we are checking only IMAGE here?

kurund: Can you also check on failing unit tests.

kurund (approved):
Looks good.

Issue #3009, opened 2024-08-07
HSM Template: Adding fallback image


akanshaaa19:
**Describe the bug**
On editing a hsm template with type as IMAGE, if there is no media returned we should show a fallback image.

**To Reproduce**
1. Go to HSM templates and find a media template
2. Click on Edit
4. There is no image shown in the simulator

kurund:
@akanshaaa19 

As discussed we can't save only the attachment type without the attachment URL. Can you please check if this is something we can handle in the backend?

Showing fallback images seems wrong to me.

PR #3054, opened 2024-09-03
Added fallback images

akanshaaa19:
target issue is #3009


kurund (commented):
please check https://github.com/glific/glific-frontend/issues/3009#issuecomment-2327491797

akanshaaa19:
we need to update the sync HSM api to create a media item to display on the edit page

PR #3130, opened 2024-11-13
File search ui feedback

akanshaaa19:
- When changing the temperature variable, then trying to save is giving this error. For any value other than 1 is giving this error. 
image

- Changed the error message for deleting an assistant

Product team feedback: https://docs.google.com/document/d/1Erw3Dc_r7OKxlndu3KKCcKqggA5JPp8d8ym3ANsX8dE/edit?tab=t.0#bookmark=id.690f0ro43qck


kurund:
@akanshaaa19 

Can you please update the issue on what you are trying to fix?

akanshaaa19:
> @akanshaaa19
> 
> Can you please update the issue on what you are trying to fix?

done

kurund (commented):
@akanshaaa19 

Looks good.

I am not sure about the change in the delete message title. I have added a comment in the feedback document. I feel we should change it back to the original as it was consistent with the other delete messages.

Let's chat

PR #3131, opened 2024-11-15
Template Flow: Removed the pre-fill name

akanshaaa19:
target issue is: #3125


kurund (approved):
@akanshaaa19 

Looks good.

The form was mostly empty hence, it was a bit difficult to understand the action.

I have made minor tweaks:
- setting title explaining the action
- set the helpful description

Issue #3132, opened 2024-11-15
Flow: No way to pin/unpin a flow from the list page


akanshaaa19:
When we are on the flows list page there is an pin icon, which is misleading for the users because they cannot unpin/pin the flow from there, we should have an option to pin/unpin the flow from list page

PR #3137, opened 2024-11-21
Added functionality to pin/unpin flows from the flows list

akanshaaa19:
target issue is #3132


kurund (changes requested):
@akanshaaa19 

Functionality is working as expected. However, refresh action works only once. Can you fix the same.

kurund (approved):
looks good.

Issue #3115, opened 2024-10-08
UI/ UX : Warning if flow not published


KrishnaDwibhashi2:
**Is your feature request related to a problem? Please describe.**
Users often forget to Publish the flow after making changes

**Describe the solution you'd like**
If a user has made any change in the flow and tries to click anywhere in the platform without publishing, Can we give a warning saying "You have made some changes in flow. Do you want to Publish before leaving?" and publish option?
This will avoid some misses.

Edge cases-
1) User might close the tab (might open a new tab to work on) without publishing

PR #3142, opened 2024-11-25
Added warning if flow is not published

akanshaaa19:
target issue is #3115


kurund (changes requested):
@akanshaaa19 

Tests are failing, can you please check

kurund (commented):
@akanshaaa19 

I don't see any warning message. 

Steps that I followed.
- created new flow and published
- tried editing the above flow and remove some node and clicked on menu links it navigated to the same without any warning.

May be I am missing something. Lets discuss during our call.

akanshaaa19:
closing because we are not picking this up right now

PR #3143, opened 2024-11-25
Autosave on clicking cancel

akanshaaa19:
ERP feedback: 
> 3) If the details in confirmation page (4th page of onboarding form) are filled and if we go back to Billing page- change any details (like billing frequency) and again come back to Comfirmation page, then the details added earlier gets vanished.

Clicking on cancel was not autosaving the data for the form


kurund (approved):
looks good

PR #3146, opened 2024-11-26
Added helptext for rejected/failed templates

akanshaaa19:
Added help text for rejected and failed templates so that the end users can know the difference between the two


kurund (changes requested):
This is a static help text and not specific to a record. Hence, I think we should just add legend on top rather than repeating it for each row.

Issue #3175, opened 2025-01-02
HSM: Template buttons not working smoothly


akanshaaa19:
**Describe the bug**
On switching template buttons the behaviour is not smooth.
Also if there is don't add body to the template and just buttons simulator message breaks.

**To Reproduce**
1. Go to HSM create page
2. Add a message
3. Click on 'Add buttons' and select phone number and add phone number 
4. Click on URL now
5. The values are still there

**Expected behavior**
It should work smoothly and values should be updated whenever something is changed

**Additional context**

PR #3176, opened 2025-01-06
Fixed behaviour of hsm buttons.

akanshaaa19:
target issue is #3175


kurund (approved):
looks good to me and fixes the reported issue.

PR #3192, opened 2025-01-16
Issues in hsm template

akanshaaa19:
1. not being able to apply for media messages
2. issues with validation-  on applying text type template without buttons it fails because of incorrect validation


vincent89291 (approved):

kurund (approved):
logic looks good. However, I was not able to replicate the issue.

PR #3206, opened 2025-02-04
Removing llm4dev code

akanshaaa19:
We are not using ll4dev now that's why will remove the related code from the frontend as well as backend


kurund (approved):
Looks good.

kurund:
@akanshaaa19 

Should we also remove it from Settings.

Screenshot 2025-02-05 at 20 57 13

Issue #3160, opened 2024-12-10
Support for rich text editor in hsm templates


akanshaaa19:
**Is your feature request related to a problem? Please describe.**
Adding rich text editor in hsm template create page so that users can format their messages.

**Describe the solution you'd like**
Like this
<img width="1115" alt="image" src="https://github.com/user-attachments/assets/5d4dbeb8-2ff5-4ed7-8c22-64664b9cb6ca">

kurund:
Note that we will be implementing the commonly used/supported formatting options only.

PR #3210, opened 2025-02-10
Added support for rich text editor

akanshaaa19:
target issue is #3160


kurund (changes requested):
@akanshaaa19 

Looks good and mostly working fine except following cases.

- Need to click emoji picker twice for the first time
- I feel we should remove option.

kurund (approved):
looks good.

Issue #3163, opened 2024-12-11
WA groups: Start a flow for group collections not possible


mahajantejas:
**Describe the bug**
If we want to "start a flow" for a "group collection" it cannot be done through the "group chats" page.

**To Reproduce**
1. Go to 'group chats'
2. Click on 'group collections'
3. find the down arrow next to the collection and try to start a flow for the WhatsApp group

**Expected behavior**
There should be an option to start the flow for a whatsapp group collection

**Screenshots**
<img width="1210" alt="Screenshot 2024-12-11 at 5 26 44 PM" src="https://github.com/user-attachments/assets/e9b4f274-ffbb-48e0-a8a2-f2ac05f49d47"/>

madclaws:
Will add an API as mentioned [here](https://github.com/glific/glific/issues/3936)

madclaws:
The api is done from backend `startWaGroupCollectionFlow`

PR #3229, opened 2025-02-24
wa groups: start a flow for collections

akanshaaa19:
target issue is #3163


kurund (approved):
Looks good. 

This is dependent on https://github.com/glific/glific/pull/4057

Issue #3108, opened 2024-09-23
Glific platform logging out suddenly


KrishnaDwibhashi2:
Glific platform (SNEHA) logged out suddenly between 4:46-4:47pm on 23rd Sept. Though there were no deployments happened at that time, Anandu suspects this will be most probably refresh token api failure.

It happened earlier with other NGOs as well- who also reported on discord

PR #3231, opened 2025-02-28
added logs for log out issues

akanshaaa19:
target issue is #3108


akanshaaa19:
@kurund tested it in scenarios when refresh token is expired it logs the user out.

kurund (approved):
looks good.

PR #3241, opened 2025-03-19
Interactive template: Handling markdown syntax in older interactive templates

akanshaaa19:
target issue is https://github.com/glific/glific/issues/3804


kurund (approved):
looks good.

Issue #2911, opened 2024-06-04
Contact History: Showing Profile


AkhileshNegi:
**Describe the bug**
When viewing contact profile for a contact that has only one profile and that profile is not an active profile for a contact the details shown at first are bit weird

**To Reproduce**
![weird](https://github.com/glific/glific-frontend/assets/40158831/45714a70-d00a-4f5c-9fd3-2b06f1be4a4f)

**Next Steps**
Based on our discussion, there can be three case for contact that need to be handled
**1. Contact has no profile:** Show option to delete the contact
**2. Contact has one profile:**  Show option to delete the profile and delete the contact
**3. Contact has more than one profile:** Show option to delete only non active profile

Also in UI we need to show the active profile if not shown already

PR #3242, opened 2025-03-19
Contact Profile: details not visible when there is only one profile and its not active

akanshaaa19:
target issue is #2911


kurund (approved):
looks good.

Issue #3187, opened 2025-01-14
"Publish" button not aligned when the number of keywords is a lot


mahajantejas:
From Saturday Art Class registration flow. 
<img width="1200" alt="Screenshot 2025-01-14 at 4 47 05 PM" src="https://github.com/user-attachments/assets/9b76ae8f-704b-483b-89f2-8d91460d0c2f" />

Vikaspal8923:
I can work on fixing this issue.

PR #3259, opened 2025-03-27
Fixed alignment of buttons when the number of keywords are a lot

akanshaaa19:
target issue is #3187

### DIFF
--- a/src/components/floweditor/FlowEditor.tsx
+++ b/src/components/floweditor/FlowEditor.tsx
@@ -148,23 +148,38 @@ export const FlowEditor = () => {
   if (flowName && flowName.flows.length > 0) {
     flowTitle = flowName.flows[0].name;
     const keywords = flowName.flows[0].keywords;
-    if (keywords.length < 8) {
-      flowKeywords = <span>{keywords.join(', ')}</span>;
-    } else {
-      flowKeywords = (
-        <span>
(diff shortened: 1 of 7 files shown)


src/components/floweditor/FlowEditor.tsx
```diff
@@ -146,7 +147,24 @@ export const FlowEditor = () => {
 
   if (flowName && flowName.flows.length > 0) {
     flowTitle = flowName.flows[0].name;
-    flowKeywords = flowName.flows[0].keywords.join(', ');
+    const keywords = flowName.flows[0].keywords;
+    if (keywords.length < 8) {
```
kurund, line 151: It might be better to restrict this based on the string length that's rendered rather than the number of words.

src/containers/Flow/FlowList/FlowList.tsx
```diff
@@ -35,7 +35,9 @@ const getName = (text: string, keywordsList: any, roles: any) => {
     <div className={styles.NameText}>
       {text}
       <br />
-      <span className={styles.Keyword}>{keywords.join(', ')}</span>
+      <span className={styles.Keyword}>
```
kurund, line 38: It might be better to restrict this based on the string length that's rendered rather than the number of words.

kurund (changes requested):

kurund (approved):
looks good.

PR #3261, opened 2025-03-27
Enhancements in Tags for Interactive Templates, HSM, and Flow List Pages

akanshaaa19:
When a new `interactive message/hsm template/ flow` is created and saved, it returns to the page where the selected tag gets cleared out. 

Enhancements made:

- If going to any creation page from tags selected in the list and then navigating back to list page- tags won't get cleared out.
- On going to creation page from tags selected in the list page- tags will be auto populated
- HSM template tags didn't work earlier

## Summary by CodeRabbit

- **New Features**
  - Tag filters for lists now synchronize with the URL, allowing direct linking and bookmarking of filtered views.
  - When creating new items from filtered lists, the selected tag is automatically passed to the creation form for a smoother workflow.
  - Navigation and back buttons dynamically reflect tag context for improved user navigation.

- **Bug Fixes**
  - Navigation and back button behavior now consistently reflect the selected tag context, improving user experience when moving between filtered lists and forms.

- **Tests**
  - Added and updated tests to cover navigation with tag filters and ensure correct state is passed during creation actions.

- **Chores**
  - Updated mock data to support new tag filtering scenarios in tests.


kurund (changes requested):
There is an issue with creating a flow using template. On save it redirects to URL like `https://flow/?isTemplate=true`

kurund (approved):
looks good.

PR #3262, opened 2025-03-31
Added codecov.yml file

akanshaaa19:


codecov.yml
```diff
@@ -0,0 +1,3 @@
+ignore:
```
kurund, line 1: Can we remove this as it's not working.

kurund (changes requested):
@akanshaaa19 

The unit tests are still failing consistently. I tried several times to re-run

PR #3328, opened 2025-05-28
Downgrade to Previous Stable Version to Fix Autocomplete TypeError on @results

akanshaaa19:
This PR downgrades the application to the last known stable version to resolve an issue with the autocomplete of results field in the floweditor.

Have also created an issue in the temba components repo: https://github.com/nyaruka/temba-components/issues/502

## Summary by CodeRabbit

- **Chores**
  - Downgraded the "@nyaruka/temba-components" dependency to an earlier version.


kurund (commented):
@akanshaaa19 

Looks like it's a big downgrade, do we know if it was broken after ^0.108.1?

If not it's worth checking which version broke this?

PR #3360, opened 2025-06-19
removed multiple logic for token refresh

### LINKED ISSUE #3348: Auth: Refactor token refresh logic to avoid conflict between Apollo and custom implementation
Our current token refresh setup is causing conflicts. We are using the apollo-link-token-refresh library, which automatically handles token renewal. However, we also have our own custom token refresh logic in place.

This results in both mechanisms trying to refresh the token simultaneously, which can lead to:
- Mixed-up tokens
- 401 unauthorized errors
- Unintended logouts

akanshaaa19:
resolves #3348 

This PR addresses issues related to duplicate token renewal calls and improves error handling across Apollo Client and the logout flow.

### Key changes:

- Removed redundant token logic from App.tsx to prevent multiple simultaneous token renewal requests, which were causing token mix-ups.
- Introduced isLoggingOut flag in apolloClient.ts to prevent retries during logout. This ensures that after a failed token renewal, retry logic doesn’t trigger unnecessary API calls.
- Simplified the onError handler by removing the status code-based switch block (since Apollo doesn't expose it reliably). Enhanced log clarity by adding gql variables to log messages.
- In the logout component previously, the app navigated immediately to the login page, causing the logout modal to flash or not appear. Now, navigate is triggered only when the user clicks the logout button, ensuring the modal is fully visible and interactive.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved logout handling to prevent multiple simultaneous logout attempts and redundant logout flows.
  - Unified logout navigation paths for consistent user experience.

- **Refactor**
  - Streamlined session validation and token renewal processes by removing redundant checks.
  - Enhanced logout flow to ensure proper navigation after logout confirmation.

- **Tests**
  - Simplified and refocused login-related tests to emphasize user interaction and rendering with mocked network responses.


madclaws:
Tested, lgtm.

src/containers/Auth/Logout/Logout.tsx
```diff
@@ -51,23 +51,27 @@ export const Logout = () => {
...
 
+  const logoutAndNavigate = () => {
+    handleLogout();
     navigate('/login', { replace: true, state: { to: location.state } });
   };
 
   useEffect(() => {
     // if user click on logout menu
     if (params.mode === 'user') {
       handleLogout();
```
kurund, line 64: May be replace this with logoutAndNavigate()

src/App.tsx
```diff
@@ -21,22 +20,6 @@ const App = () => {
   const [drawerOpen, setDrawerOpen] = useState(true);
   const isAuthenticated = !!getAuthSession('accessToken');
```
kurund, line 21: I am wondering if this needs to be inside useEffect()

src/config/apolloclient.ts
```diff
@@ -63,6 +64,10 @@ const gqlClient = (navigate: any) => {
       return tokenResponse;
     },
     handleError: (err: Error) => {
+      // Prevent multiple logout attempts
```
kurund, line 67: why would this happen if we are navigating away?

src/config/apolloclient.ts
```diff
@@ -100,26 +103,18 @@ const gqlClient = (navigate: any) => {
...
-          checkAndRenewToken(navigate);
-          break;
-        default:
-          // eslint-disable-next-line
-          Sentry.captureException(`[Network error]: ${networkError}`);
-          // logged error in logflare
-          setLogs(`[Network error]: ${networkError}`, 'error');
-      }
+      setLogs(`Network error: ${networkError} ${operation.variables}`, 'error');
+      navigate('/logout/session', { state: { to: window.location.pathname } });
```
kurund, line 107: Does this include a temporary loss of internet? If so, then I am not sure if we should logout.

src/config/apolloclient.ts
```diff
@@ -100,26 +103,18 @@ const gqlClient = (navigate: any) => {
...
     }
   });
 
   const httpLink: any = createLink({ uri: GLIFIC_API_URL });
 
   const retryIf = (error: any) => {
+    if (isLoggingOut) {
```
kurund, line 114: same as above.

kurund (changes requested):

src/config/apolloclient.ts
akanshaaa19: it will cause login delays and we might come across the issue where we had to login twice

kurund: Why would that happen? Can you give provide steps to replicate?

src/App.tsx
kurund: @akanshaaa19 

Did you check on this?

kurund (approved):
looks good

PR #3367, opened 2025-06-24
Profiles: Is_active filter in the profile list

AmishaBisht:


package.json
```diff
@@ -132,4 +132,4 @@
     "vitest": "^3.1.3",
     "yarn-upgrade-all": "^0.7.5"
   }
-}
+}
```
shijithkjayan, line 135: **issue:** EOF error

src/graphql/queries/Contact.ts
```diff
@@ -168,6 +168,7 @@ export const GET_CONTACT_PROFILES = gql`
       id
       name
       type
+      is_active
```
shijithkjayan, line 171: **suggestion:** Please also add `is_default`

akanshaaa19 (changes requested):
image
not able to see any profiles here

PR #3369, opened 2025-06-25
Trim leading spaces from media URL

priyanshu6238:
**Summary**

This PR ensures that leading and trailing whitespace in media URLs are trimmed before validation. Previously, users encountered an “Invalid media URL” error when pasting URLs with accidental spaces — even though the URLs were otherwise correct. This led to unnecessary confusion and validation failures.

**Motivation**

While creating templates, users often paste media URLs that may unintentionally include a leading or trailing space. Since the backend validation treats such URLs as invalid, it results in a poor user experience. By trimming the URL before calling validateMedia, we eliminate this friction and improve form reliability.

**Changes Made**
- Trimmed the media URL using value.trim() before passing it to the validateMedia function.
- Ensured no changes to other validation logic or UI.

**Test Plan**

- Pasted a URL with leading/trailing spaces → It now passes validation if the actual URL is valid.
- Pasted a completely invalid URL → Still fails as expected.
- Confirmed that trimming doesn't affect already valid URLs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved URL handling by trimming whitespace from attachment URLs before validation.
  - Enhanced form validation to prevent errors when URLs contain leading or trailing spaces.
  - Updated attachment URL input behavior to validate and save trimmed values only on input blur, reducing unnecessary state updates.

### DIFF
--- a/src/containers/HSM/HSM.test.tsx
+++ b/src/containers/HSM/HSM.test.tsx
@@ -33,7 +33,6 @@ vi.mock('lexical-beautiful-mentions', async (importOriginal) => {
   };
 });
 
-// Add this spy before your tests
 const validateMediaSpy = vi.spyOn(utilsModule, 'validateMedia');
 
 describe('Edit mode', () => {
--- a/src/containers/HSM/HSM.test.tsx
+++ b/src/containers/HSM/HSM.test.tsx
@@ -5,6 +5,7 @@ import { MemoryRouter, Route, Routes } from 'react-router';
 import { HSM } from './HSM';
 import { HSM_TEMPLATE_MOCKS, getHSMTemplateTypeMedia, getHSMTemplateTypeText } from 'mocks/Template';
 import { setNotification } from 'common/notification';
+import * as utilsModule from 'common/utils';
 
 const mocks = HSM_TEMPLATE_MOCKS;
 
@@ -31,6 +32,9 @@ vi.mock('lexical-beautiful-mentions', async (importOriginal) => {
     BeautifulMentionsMenuItemProps: {},
   };
 });
+
+// Add this spy before your tests
+const validateMediaSpy = vi.spyOn(utilsModule, 'validateMedia');
 
 describe('Edit mode', () => {
   test('HSM form is loaded correctly in edit mode', async () => {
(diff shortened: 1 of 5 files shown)


akanshaaa19 (changes requested):
Can we apply this same logic across all features that deal with attachments(like interactive messages, speed sends, etc.)? 
Maybe we can also extract this as a shared function so we can reuse it consistently and avoid duplication.

akanshaaa19:
Also add test cases for this.

src/containers/HSM/HSM.tsx
```diff
@@ -323,9 +323,10 @@ export const HSM = () => {
   };
 
   const validateURL = (value: string) => {
-    if (value && type) {
+    const trimmedValue = value.trim();
```
shijithkjayan, line 326: wont this throw errors when `value` has no value, ie; undefined?

src/containers/HSM/HSM.tsx
```diff
@@ -322,19 +322,21 @@ export const HSM = () => {
...
-        } else {
-          setIsUrlValid('');
-        }
-        setValidatingURL(false);
-      });
-    }
-  };
+const validateURL = (value: string) => {
+  if (value && type) {
+    // const trimmedValue = value.trim();
```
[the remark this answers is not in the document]
akanshaaa19: That's okay but can we should still remove the commented out line.

src/containers/HSM/HSM.tsx
```diff
@@ -322,19 +322,21 @@ export const HSM = () => {
...
-        if (!response.data.is_valid) {
-          setIsUrlValid(response.data.message);
-        } else {
-          setIsUrlValid('');
-        }
-        setValidatingURL(false);
-      });
-    }
-  };
+const validateURL = (value: string) => {
```
akanshaaa19, line 325: The formatting is messed up here. Can you make sure you have "format on save" enabled and prettier installed in your IDE

src/containers/HSM/HSM.tsx
```diff
@@ -587,11 +589,11 @@ export const HSM = () => {
       ),
       inputProp: {
         onBlur: (event: any) => {
-          setAttachmentURL(event.target.value);
+          setAttachmentURL(event.target.value.trim());
         },
         onChange: (event: any) => {
           clearTimeout(timer);
-          timer = setTimeout(() => setAttachmentURL(event.target.value), 1000);
+          timer = setTimeout(() => setAttachmentURL(event.target.value.trim()), 1000);
```
akanshaaa19, line 596: Can you check why do we have both onBlur and onChange, i think we can remove onChange from here. Can you check if that is breaking anything

src/containers/HSM/HSM.tsx
```diff
@@ -587,11 +589,11 @@ export const HSM = () => {
       ),
       inputProp: {
         onBlur: (event: any) => {
-          setAttachmentURL(event.target.value);
+          setAttachmentURL(event.target.value.trim());
```
akanshaaa19, line 592: This should be added everywhere- Interactive messages, Speed sends or anywhere else where we are validating a URL and sending it to the BE

akanshaaa19 (changes requested):

src/containers/HSM/HSM.tsx
priyanshu6238: Yes, we can remove it

src/containers/HSM/HSM.test.tsx
```diff
@@ -335,4 +335,35 @@ describe('Add mode', () => {
...
+      // The input value should still have spaces, but validation should use the trimmed value.
+      expect(inputs[3]).toHaveValue('   https://example.com/image.jpg   ');
+    });
+
+    // Optionally, trigger blur to invoke validation
+    fireEvent.blur(inputs[3]);
+
+    // There should be no validation error for a valid trimmed URL
+    await waitFor(() => {
+      expect(screen.queryByText('Attachment URL is required.')).not.toBeInTheDocument();
```
akanshaaa19, line 366: We are testing for invalid URL. 
this `expect(screen.queryByText('Attachment URL is required.')).not.toBeInTheDocument();` is always true because the URL is present.
Change this to test if the validateUrl func is called correctly with no spaces.

src/containers/HSM/HSM.tsx
```diff
@@ -322,19 +322,19 @@ export const HSM = () => {
...
+    validateMedia(value, type.id, false).then((response: any) => {
+      if (!response.data.is_valid) {
+        setIsUrlValid(response.data.message);
+      } else {
+        setIsUrlValid('');
+      }
+      setValidatingURL(false);
+    });
+  }
+};
```
akanshaaa19, line 337: Can you please fix the formatting here?

src/containers/SpeedSend/SpeedSend.test.tsx
```diff
@@ -120,7 +119,7 @@ describe('test creating a speed send', () => {
 
     fireEvent.click(screen.getByText('IMAGE'), { key: 'Enter' });
 
-    fireEvent.change(inputs[2], {
+    fireEvent.blur(inputs[2], {
```
akanshaaa19, line 122: ideally we should first enter the URL then blur so that it is more understandable.

akanshaaa19 (changes requested):

akanshaaa19 (changes requested):

akanshaaa19 (approved):

Issue #3372, opened 2025-06-26
Close button not working


shijithkjayan:
**Describe the bug**
The close button in the google sheet error pop up is not working

**To Reproduce**
1. Go to [Sheets](https://prod.glific.com/sheet-integration)
2. Click on sync button for the sheet named `Failing Google Sheet`
4. In the pop up that appears, click on the close button at the top

**Expected behavior**
Pop up should close

**Screenshots**
https://github.com/user-attachments/assets/9ddef645-fac6-4a2a-a9d3-7400373ce873

PR #3391, opened 2025-07-12
Fix dialog close behavior

priyanshu6238:
## Summary
 Target issue is https://github.com/glific/glific-frontend/issues/3372

- Fixed the issue where the "Close"  icon in the warning dialog was not working as expected.
- The root cause was that the handleCancel prop was not being passed to the DialogBox component, causing the close icon and backdrop click to have no effect.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to close the warning dialog using the close (cross) icon in the Sheet Integration List.

* **Tests**
  * Added a test to verify that the warning dialog closes when the close icon is clicked.


akanshaaa19 (changes requested):
@priyanshu6238 can you add or modify existing test case covering the changed line?

akanshaaa19 (approved):

Issue #3250, opened 2025-03-24
UI/UX: HSM: enable adding a footer in the HSM templates


mahajantejas:
There is an optional footer field, which can be useful for NGOs to use. 
HSM templates with footer cannot be applied from Glific templates page because the footer field does not exist. 

<img width="1204" alt="Image" src="https://github.com/user-attachments/assets/aef69672-01eb-4f28-9384-c3ea2118580a" />

PR #3397, opened 2025-07-16
Add footer field to HSM component with validation

priyanshu6238:
## Summary
 Target issue is https://github.com/glific/glific-frontend/issues/3250
### Motivation

-  Gupshup supports an optional footer field in HSM templates for additional message context.
- Glific currently does not support the footer, limiting the ability to utilize this feature.

## Summary by CodeRabbit

* **New Features**
  * Added support for an optional footer field in template messages, allowing users to add and preview a footer in the message simulator.
  * The footer field is now included in template creation, editing, and display.

* **Style**
  * Updated simulator styles to support the new footer field and improved input focus and hover effects.

* **Bug Fixes**
  * Disabled the "Add" button in template options when appropriate.

* **Tests**
  * Updated tests to cover the new footer field in template messages.

* **Chores**
  * Enhanced mock data and GraphQL operations to support the footer field.


src/containers/HSM/HSM.tsx
```diff
@@ -503,6 +516,32 @@ export const HSM = () => {
...
+          }
+
+          const parsedText = parse.length ? `| ${parse.join(' | ')}` : '';
+          if (parsedText) {
+            updatedMessage += parsedText;
+          }
+
+          setSimulatorMessage(updatedMessage);
+        },
+      },
```
akanshaaa19, line 543: why is this added here?

src/containers/HSM/HSM.tsx
```diff
@@ -408,17 +415,23 @@ export const HSM = () => {
...
-    let typeValue;
-    let text = message;
+    let typeValue = type?.id || 'TEXT';
 
     mediaBody.caption = getExampleFromBody(body, variables);
     mediaBody.url = attachmentURL;
-    typeValue = type?.id || 'TEXT';
 
-    setSampleMessages({ ...sampleMessages, body: text, media: mediaBody, type: typeValue });
+    const fullMessage = footerText ? `${trimmedBody}\n${footerText}` : trimmedBody;
```
akanshaaa19, line 427: footer is not a part of the message body it should be a separate element, as the styling also differs from the  message body. instead of this you should add a footer field in the sampleMessages object and then update that when footer is updated and also add another footer element in simulator.tsx from where we render the HSM message.

akanshaaa19 (changes requested):

src/containers/HSM/HSM.tsx
```diff
@@ -263,7 +268,15 @@ export const HSM = () => {
...
-      setSimulatorMessage(getExampleFromBody(bodyValue, variables));
+      const { message } = getTemplateAndButton(getExampleFromBody(bodyValue, variables));
+
+      setSampleMessages({
+        type: typeValue?.id || 'TEXT',
+        body: message,
+        media: { ...sampleMessages.media },
+        footer: footerValue || '',
+        location: null,
+      });
```
akanshaaa19, line 279: we have the setSimulatorMessage function so that we don't have to make changes to sampleMessages again and again and to maintain consistency. It should not be replaced.

akanshaaa19 (changes requested):

akanshaaa19 (changes requested):
looks good, just add localization for the validation.

akanshaaa19 (approved):

Issue #3426, opened 2025-08-07
Quick onboarding: Show user email on the account page


madclaws:
**Describe the task**
Since we are collecting user email while signup, we can show that in the `/myaccount` page

PR #3430, opened 2025-08-13
Add email field to my account page

priyanshu6238:
## Summary
 Target issue - https://github.com/glific/glific-frontend/issues/3426
Use the backend branch feat/quick_onboarding to test this PR.

## Summary by CodeRabbit

* New Features
  * My Account now displays the user’s email address as a read-only field alongside name and phone.
  * Form initialization updated to include email for a consistent profile overview.

* Localization
  * Added English translation for the Email label.

### DIFF
--- a/.github/workflows/cypress-testing.yml
+++ b/.github/workflows/cypress-testing.yml
@@ -65,7 +65,6 @@ jobs:
           git clone https://github.com/glific/glific.git
           echo done. go to dir.
           cd glific
-          git checkout feat/quick_onboarding
           echo done. start dev.secret.exs config
           cd priv
           mkdir cert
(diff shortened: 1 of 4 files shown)


akanshaaa19:
can you look at why codacy is failing @priyanshu6238

akanshaaa19:
> The entire accounts form is non editable, do we want to make email editable 
imo lets make email editable, that way older orgs can also update their email if they want

madclaws:
> imo lets make email editable, that way older orgs can also update their email if they want

But this will only update user account's email rather than org's email. The point of collection email of user on signup is to later use it in email groups etc. Do we need to spend time on that now?

akanshaaa19:
@madclaws If backend doesn’t support it right now, let’s leave it non-editable. We can pick it up later, but let’s create an issue to for this. @priyanshu6238

akanshaaa19 (approved):

shijithkjayan:
@KrishnaDwibhashi2 yes, the updated email ID will replace the user's existing email ID in the database.

For now let's keep it as non-editable, and when we work on phase 2 of quick onboarding we can make the field editable.

Issue #3423, opened 2025-08-07
Quick onboarding - Enhancements for one-time bot number updation


madclaws:
**Describe the task**

UI enhancements for one-time bot number update

- [ ] Add a confirmation popup that it would not be possible to update number later.
- [ ] If `allow_bot_number_update` in settings is true, then phnum text block is editable, else no.

shijithkjayan:
@madclaws Lets also show the number in the pop up form for a better UX, so even if the user wants to verify the details they don't have to close the pop. cc: @KrishnaDwibhashi2

PR #3445, opened 2025-08-26
Quick Onboarding: Made org phone number one-time editable

madclaws:
Target issue https://github.com/glific/glific-frontend/issues/3423

BE branch: `feat/quick_onboarding`

### DIFF
--- a/src/containers/SettingList/Organization/Organization.tsx
+++ b/src/containers/SettingList/Organization/Organization.tsx
@@ -142,7 +142,7 @@ export const Organization = () => {
   };
 
   const FormSchema = Yup.object().shape(validation);
-  const allowBotNumberUpdate = orgData?.organization?.organization?.setting?.allow_bot_number_update;
+  const allowBotNumberUpdate = orgData?.organization?.organization?.setting?.allowBotNumberUpdate;
   const formFields: any = [
     {
       component: Input,
@@ -182,6 +182,7 @@ export const Organization = () => {
       type: 'text',
       label: t('Organization phone number'),
       disabled: !allowBotNumberUpdate,
+      onChange: (e: string) => setPhone(e),
       endAdornment: (
         <InputAdornment position="end">
           <IconButton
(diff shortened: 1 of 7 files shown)


akanshaaa19:
can you also include the backend branch name to test this in the description? @madclaws

madclaws:
> can you also include the backend branch name to test this in the description? @madclaws

done

src/containers/SettingList/Organization/Organization.tsx
```diff
@@ -128,6 +128,7 @@ export const Organization = () => {
 
   const validation = {
     name: Yup.string().required(t('Organization name is required.')),
+    phone: Yup.string().required(t('Phone number is required.')),
```
akanshaaa19, line 131: we can add a minimum and maximum length validation for the phone number here

src/containers/SettingList/Organization/Organization.tsx
```diff
@@ -180,7 +181,8 @@ export const Organization = () => {
       name: 'phone',
       type: 'text',
       label: t('Organization phone number'),
-      disabled: true,
+      disabled: !allowBotNumberUpdate,
+      onChange: (e: string) => setPhone(e),
       endAdornment: (
         <InputAdornment position="end">
           <IconButton
```
akanshaaa19, line 188: if there’s no phone number, we can just hide the endAdornment.

akanshaaa19 (changes requested):
looks good to me, only two minor changes needed.

src/containers/SettingList/Organization/Organization.tsx
madclaws: We have a strong E.164 standard validation on BE which covers all the case, adding min and max on FE doesn't add much value and is just a placeholder. However i added a max 30 so that users can't abuse the system by sending huge amount of data in phnum textbox.

akanshaaa19 (approved):

Issue #3501, opened 2025-09-08
Quick Onboarding Feedback: Improve Field Order in Gupshup Page (Settings)


priyanshu6238:
## **Description:**
On the Gupshup page inside Settings, the current field order is not intuitive. The App ID field (non-editable) is positioned between two editable fields (App Name and API Key).

This interrupts the natural flow for users when entering or reviewing credentials.
### **Suggested Improvement:**
Rearrange the fields so that the API Key appears above the App ID, ensuring both editable fields (App Name and API Key) are grouped together, followed by the non-editable App ID.

This adjustment will improve usability and provide a smoother onboarding experience.

PR #3507, opened 2025-09-08
Quick Onboarding: Improve field ordering in gupshup and registration page

priyanshu6238:
Target issue: https://github.com/glific/glific-frontend/issues/3501

### DIFF
--- a/src/containers/Auth/Registration/Registration.tsx
+++ b/src/containers/Auth/Registration/Registration.tsx
@@ -58,6 +58,14 @@ export const Registration = () => {
       autoComplete: 'name',
     },
     {
+      component: Input,
+      name: 'email',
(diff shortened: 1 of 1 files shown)


src/containers/SettingList/Providers/Providers.tsx
```diff
@@ -134,23 +134,43 @@ export const Providers = () => {
...
+            disabled: fields[key].view_only,
+            skip: fields[key].hide,
+          };
+          formField.push(field);
+
+          // create validation object for field
+          addValidation(fields, key);
+
+          // add default value for the field
+          states[key] = fields[key].default || '';
```
madclaws, line 155: There's some code duplication, can we see if we can refactor and reuse common parts.

Issue #3524, opened 2025-09-22
API Payload Update for Import Contacts:


priyanshu6238:
The id field is no longer required in the request payload. The backend automatically determines the organization based on the authenticated user, so the frontend does not need to send the ID when importing contacts.

<img width="413" height="297" alt="Image" src="https://github.com/user-attachments/assets/33038440-7ddc-48e8-9b34-07daad11cc4b" />

PR #3526, opened 2025-09-23
Remove unused parameter from ImportContacts Api payload

priyanshu6238:
## Summary

Target issue: https://github.com/glific/glific-frontend/issues/3524

## Summary by CodeRabbit

* **Refactor**
  * Contact import and move flows simplified: the app no longer requires or sends an organization ID when uploading or moving contacts.
  * Backend mutation payloads streamlined to omit the ID parameter; upload dialog now submits only the data file, optional group label, and import type.
  * Displayed results, status indicators, and error messages remain unchanged.

### DIFF
--- a/src/containers/ContactManagement/UploadContactsDialog/UploadContactsDialog.tsx
+++ b/src/containers/ContactManagement/UploadContactsDialog/UploadContactsDialog.tsx
@@ -10,7 +10,6 @@ import { Checkbox } from 'components/UI/Form/Checkbox/Checkbox';
 import { Field, Form, Formik } from 'formik';
 import { UPLOAD_CONTACTS_SAMPLE } from 'config';
 import { IMPORT_CONTACTS } from 'graphql/mutations/Contact';
-import { getUserSession } from 'services/AuthService';
 import { setNotification } from 'common/notification';
 import styles from './UploadContactsDialog.module.css';
 import UploadIcon from 'assets/images/icons/UploadIcon.svg?react';


madclaws:
@priyanshu6238 Can we remove passing `id` in the api request 
https://github.com/glific/glific-frontend/blob/930d7d676d4611f0a443b54f13ed2805dd531305/src/containers/ContactManagement/UploadContactsDialog/UploadContactsDialog.tsx#L96

madclaws (approved):

src/graphql/mutations/Contact.ts
shijithkjayan, line 1: Shouldn't we also remove the ID from the move contacts query?

src/containers/ContactManagement/UploadContactsDialog/UploadContactsDialog.tsx
shijithkjayan, line 1: Don't we have to remove the ID from the move contacts API here also?

priyanshu6238: In the frontend, we never use the organization ID in the move_contact payload, so no changes are needed here. For example ,In src/containers/ContactManagement/AdminContactManagement/AdminContactManagement.tsx, if you go to the last section, you will  see that the payload only  includes type and data. 
Screenshot 2025-09-29 at 12 04 07 PM

shijithkjayan (approved):

Issue #3533, opened 2025-09-26
Failed to verify Google Captcha issue


priyanshu6238:
## **Description**

Currently, in our Captcha component, the reCAPTCHA token is generated immediately when the page loads. This causes a problem: if the user waits for more than 2 minutes (the typical token validity period) before submitting the registration form, the token expires. As a result, the backend verification fails with the error “failed to verify the captcha.”

###  **Expected Behavior**

1. A valid token should always be available when the user submits the form.
2. Token should not expire before form submission, regardless of how long the user spends on the page.

PR #3534, opened 2025-09-26
Regenerate reCAPTCHA token on form submit to prevent expiry on page load

priyanshu6238:
## **Summary**

Target issue: https://github.com/glific/glific-frontend/issues/3533

### DIFF
--- a/src/components/UI/Form/Captcha/Captcha.tsx
+++ b/src/components/UI/Form/Captcha/Captcha.tsx
@@ -2,7 +2,6 @@ import { useEffect, useCallback } from 'react';
 import { useGoogleReCaptcha } from 'react-google-recaptcha-v3';
 
 export interface CaptchaProps {
-  onTokenUpdate: any;
   action: string;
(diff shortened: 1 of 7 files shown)


src/components/UI/Form/Captcha/Captcha.tsx
```diff
@@ -1,36 +1,35 @@
...
-  }, [handleReCaptchaVerify]);
+    return token;
+  }, [executeRecaptcha]);
 
   return (
     <Component
-      onClick={() => {
-        handleReCaptchaVerify().then(() => onClick());
+      onClick={async () => {
+        const token = await handleReCaptchaVerify();
```
akanshaaa19, line 31: should add an if condition here to check if token is available or not

akanshaaa19 (changes requested):
please make the necessary changes to the onboarding form also and remove the unused setFieldValue function from here: `/src/containers/Auth/Auth.tsx:164`

PR #3553, opened 2025-10-07
Refactor Flow Editor auth handling: remove setAuthHeader and improve token renewal

akanshaaa19:
- Removed setAuthHeader function
- Added a 15-min health check to keep tokens fresh.
- Tokens now renew 1 minute before expiry to avoid failures in floweditor.

## Summary by CodeRabbit

- New Features
  - Automatically keeps sessions active during flow editing with periodic health checks.
- Bug Fixes
  - Reduced unexpected sign-outs by adding a safety buffer before token expiry.
- Refactor
  - Removed legacy background request behaviors to improve stability and reduce side effects.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "@emoji-mart/react": "^1.1.1",
     "@emotion/react": "^11.14.0",
     "@emotion/styled": "^11.14.1",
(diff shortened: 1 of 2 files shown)


akanshaaa19:
we found an alternative solution for this issue, will get feedback if that is working and then investigate it further

Issue #3491, opened 2025-08-29
Have upload attachment option in Interactive message page


KrishnaDwibhashi2:
Have upload from system attachment option in Interactive message page similar to Attachments tab of "Send contact a message node"

PR #3594, opened 2025-10-29
Interactive Message: add upload attachment option

AmishaBisht:
Target issue: #3491

## Summary by CodeRabbit

* **New Features**
  * File upload added to interactive messages (image, video, PDF).
  * New "Upload attachment" option in attachment types to start uploads.
  * Automatic message type assignment based on chosen attachment category.
  * Real-time upload status with uploaded file name shown.
  * Attachment URL input is disabled during active uploads to prevent conflicts.


priyanshu6238 (commented):
When we select the Attachment type and upload a file, the Attachment URL field becomes non-editable.
Image

akanshaaa19 (dismissed):
can we have an upload button instead of abruptly opening the file upload dialog on selecting upload attachment. it seems buggy

akanshaaa19 (approved):

Issue #3596, opened 2025-08-26
Approved HSM element name not immediately available in dropdown for translation


KrishnaDwibhashi2:
When submitting a new HSM message as a translation of an existing one, you need to select the original HSM from the Element Name dropdown. However, if an HSM is approved in one language (e.g., English), it doesn’t appear immediately in the dropdown when applying for another language. Even after refreshing, the list isn’t updated quickly.Because of this delay, you cannot submit translations right after the first approval and need to wait before submitting in multiple languages.

Steps to Replicate- 
1) Go to the HSM page.
2) Submit a new HSM message in 1 language
3) Post approval, try submitting translation of the already approved HSM.
4) In the Element Name dropdown, search for the recently approved HSM.
5) Notice that the newly approved HSM does not appear immediately in the dropdown.Even after refreshing the page, the list takes time to update.

priyanshu6238:
closing wrt https://github.com/glific/glific-frontend/pull/3648

PR #3597, opened 2025-10-29
HSM: Deduped the hsm shortcodes for Autocomplete

madclaws:
Target issue https://github.com/glific/glific-frontend/issues/3596

If autocomplete component has duplicate elements, its acting weird by showing the duplicates in the dropdown even if those elements are not in search query. Due to this the actual searched element is deep down in the list, so the users miss it sometimes thinking its not in the list

> Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.

Error from autocomplete.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session template handling to prevent duplicates and invalid entries from appearing.

### DIFF
--- a/src/containers/HSM/HSM.tsx
+++ b/src/containers/HSM/HSM.tsx
@@ -127,12 +127,12 @@ export const HSM = () => {
 
   const shortCodeOptions: any = [];
   if (shortCodes?.sessionTemplates) {
-    const seen = new Set<string>();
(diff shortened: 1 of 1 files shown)


akanshaaa19 (approved):
looks good but can we use better naming for the variables

PR #3745, opened 2026-01-30
Fix Vite build memory exhaustion on Gigalixir

akanshaaa19:
This PR fixes Vite build failures due to JavaScript heap out-of-memory errors on Gigalixir by reducing peak memory usage during the production build.

The changes focus on build-time optimizations only and do not affect runtime behavior.

## Summary by CodeRabbit

* **Chores**
  * Build now conditionally enables error-tracking integration only when credentials are present.
  * Source map generation is tied to the error-tracking setting to reduce unnecessary artifacts.
  * Improved dependency handling and manual bundle splitting (React, GraphQL, UI libs, vendor groups) to speed application load and builds.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


mdshamoon (approved):
We might want to explicitly set memory limits for node. Otherwise looks good.

PR #3756, opened 2026-02-07
Optimize Cypress

priyanshu6238:
## Summary by CodeRabbit

* **Tests**
  * Enabled CI test sharding into 3 shards with visible shard progress labels, disabled fail-fast so all shards complete, and configured split indexing for parallel Cypress runs.
* **Chores**
  * Added a test-splitting dependency and set environment variables to coordinate Cypress test distribution across shards.


mdshamoon (approved):
This is great. Is 3 the maximum no. of shards that we can have?

Issue #3727, opened 2026-01-15
Unified APIs: Block editing legacy vector stores


shijithkjayan:
### Summary

Unified APIs requires access to the files uploaded to a vector store. But for existing vector stores, the files were uploaded to OpenAI and we don't have access to these files. So we should block updating of legacy vector stores.

### Requirements

- `Manage Files` button should be greyed out in assistants show page when the vector store is marked as legacy.
- Should show a message on hovering over the button why the button is greyed out.

### Acceptance Criteria

- Button should be greyed out
- Reason for greying out should be shown

PR #3757, opened 2026-02-08
Unified APIs: Block editing legacy vector stores

priyanshu6238:
## Summary

Target issue: https://github.com/glific/glific-frontend/issues/3727
- Added legacy field to the GraphQL query                                                                                                                                 
- Disabled the Manage Files button when legacy is true and wrapped it with an MUI Tooltip that shows This vector   store was created  externally and cannot be modified here. on hover, explaining why the button is greyed out.                                             
-  Added a test case that verifies the button is disabled and the tooltip message appears for legacy vector stores


priyanshu6238:
Closing these as Cypress changes are also required. Everything will be handled in this PR: https://github.com/glific/glific-frontend/pull/3758

PR #3766, opened 2026-02-12
Unified API: Show indicator for unsaved changes

akanshaaa19:


src/containers/Assistants/CreateAssistant/CreateAssistant.tsx
priyanshu6238: Image

I accidentally added an extra space in the instruction, and it’s now showing as unsaved changes. For a few minutes, I couldn’t figure out what the unsaved change was.

Can we trim the value before checking to avoid this issue?

priyanshu6238 (approved):

mdshamoon (approved):

PR #3796, opened 2026-03-01
Fix Unified API Bugs

shijithkjayan:
## Summary

Fixes #4818 — Bug 4 and Bug 6.

### Bug 6: "Knowledge base required" warning persists after file upload

The warning was rendered based on `formikErrors?.knowledgeBaseId`, which relies on Formik's async Yup re-validation after `setFieldValue`. Changed the condition to check `!formikValues.knowledgeBaseId` directly so the warning:

- **Appears** when the user submits without a knowledge base
- **Clears immediately** when the knowledge base is created (i.e. when the `createKnowledgeBase` mutation completes and `knowledgeBaseId` is set in form state)
- **Stays visible** if files are uploaded to the dialog but not yet saved

```diff
- {formikTouched?.knowledgeBaseId && formikErrors?.knowledgeBaseId && (
+ {formikTouched?.knowledgeBaseId && !formikValues.knowledgeBaseId && (
    <p className={styles.ErrorText}>{formikErrors.knowledgeBaseId}</p>
  )}
```

### Bug 4: File size exceeded notification shown as error instead of warning

The "File size should be less than 20MB" toast was using `'error'` severity. Changed to `'warning'` since the user can simply not upload that file — it is not a blocking error.

```diff
- setNotification('File size should be less than 20MB', 'error');
+ setNotification('File size should be less than 20MB', 'warning');
```

### Other

- Cypress workflow updated to check out `master` instead of `feat/unified_APIs_2`

## Test Plan

Added two tests to `Assistants.test.tsx`:

- **Warning appears on submit without KB** — verifies the validation message is shown when the user submits the create form without a knowledge base
- **Warning clears after KB creation** — verifies the message disappears after `createKnowledgeBase` completes, not just after files are uploaded to the dialog

## Summary by CodeRabbit

* **Bug Fixes**
  * File size validation for files >20MB now shows a warning instead of an error.
  * Knowledge base ID validation message now appears only when the field is touched and left empty.

* **Tests**
  * Added tests covering assistant creation flow, knowledge base warnings, and file upload/confirmation behavior.

### DIFF
--- a/.github/workflows/cypress-testing.yml
+++ b/.github/workflows/cypress-testing.yml
@@ -65,7 +65,6 @@ jobs:
           git clone https://github.com/glific/glific.git
           echo done. go to dir.
           cd glific
-          git checkout master
           echo done. start dev.secret.exs config
           cd priv
           mkdir cert
@@ -97,7 +96,6 @@ jobs:
           git clone https://github.com/glific/cypress-testing.git
           echo done. go to dir.
           cd cypress-testing
-          git checkout main
           cd ..
           cp -r cypress-testing/cypress cypress
           yarn add cypress@13.6.2


src/containers/Assistants/Assistants.test.tsx
priyanshu6238: This test case is failing can we pls fix this

priyanshu6238 (approved):

PR #3803, opened 2026-03-03
fix: make Gupshup credential fields editable until all three are saved

shijithkjayan:
## Summary

- Removed the blanket read-only condition that locked all Gupshup fields whenever `app_id` had a non-`NA` value
- Each field (`app_name`, `api_key`, `app_id`) is now individually editable until it has a saved value other than `NA`
- The Save button is locked only when **all three** fields have a saved value (not just `app_id`)
- Added placeholder text and required validation to `app_id` (consistent with the other two fields)
- Added `app_id` to the save confirmation dialog
- Extracted the field list to a module-level constant `GUPSHUP_CREDENTIAL_FIELDS` to avoid repetition

Also removed the promotion banner from login page.

## Test Plan

- [ ] Open Gupshup settings with no credentials saved — all three fields should be editable
- [ ] Save all three fields with real values — all fields become read-only and Save button shows "Credentials Locked"
- [ ] Verify `app_id` shows placeholder text and triggers required validation when left empty

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Credential validation now treats all required fields consistently and fixes completeness checks for provider settings.

* **Improvements**
  * Confirmation dialogs now include the App ID alongside other credential details.
  * Streamlined provider settings form behavior, field ordering, placeholders and input visibility for a more consistent UI.
  * Removed promotional element from the login screen.

* **Localization**
  * Added translation for the "App ID" label.

### DIFF
--- a/src/containers/SettingList/Providers/Providers.tsx
+++ b/src/containers/SettingList/Providers/Providers.tsx
@@ -191,9 +191,10 @@ export const Providers = () => {
           : {};
 
         if (type === 'gupshup') {
+          const allFieldsSet = areAllGupshupFieldsSet(credentials);
           Object.keys(fields).forEach((key) => {
             if (GUPSHUP_CREDENTIAL_FIELDS.includes(key)) {
-              fields[key].view_only = !!(credentials[key] && credentials[key] !== 'NA');
+              fields[key].view_only = allFieldsSet;
             }
           });
           setIsDisabled(areAllGupshupFieldsSet(credentials));


src/containers/Auth/Auth.tsx
```diff
@@ -15,7 +15,6 @@ import setLogs from 'config/logs';
 import { checkOrgStatus } from 'services/AuthService';
 import { TERMS_OF_USE_LINK } from 'common/constants';
 
-import { Promotion } from './Promotion/Promotion';
```
priyanshu6238: Would it be possible to comment out this line instead of removing it? It might be useful for future reference.

src/i18n/en/en.json
```diff
@@ -349,6 +349,7 @@
   "Link": "Link",
   "App Name": "App Name",
   "API Key": "API Key",
+  "API ID": "API ID",
```
priyanshu6238:
```suggestion
  "App ID": "App ID", 
```

shijithkjayan:
> > > I am unable to enter my app_name and API key, it seems that these fields are disabled.
> > 
> > 
> > @priyanshu6238 Can you check whats the value in your credentials table?
> 
> Screenshot 2026-03-05 at 7 20 33 PM

@priyanshu6238 You have values for all three of the keys, thats why its disabled

PR #3843, opened 2026-03-15
Remove read only condition on app

shijithkjayan:
## Summary

- Removed the blanket read-only condition that locked all Gupshup fields whenever `app_id` had a non-`NA` value
- Each field (`app_name`, `api_key`, `app_id`) is now individually editable until it has a saved value other than `NA`
- The Save button is locked only when **all three** fields have a saved value (not just `app_id`)
- Added placeholder text and required validation to `app_id` (consistent with the other two fields)
- Added `app_id` to the save confirmation dialog
- Extracted the field list to a module-level constant `GUPSHUP_CREDENTIAL_FIELDS` to avoid repetition

Also removed the promotion banner from login page.

## Test Plan

- [ ] Open Gupshup settings with no credentials saved — all three fields should be editable
- [ ] Save all three fields with real values — all fields become read-only and Save button shows "Credentials Locked"
- [ ] Verify `app_id` shows placeholder text and triggers required validation when left empty

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **Improvements**
  * Standardized Gupshup credential handling so all required fields (App Name, API Key, App ID) are evaluated consistently; view-only and disabled states follow complete-field presence.
  * Credential saves now force Gupshup credentials to be active.
  * Confirmation dialog now displays App ID alongside App Name and API Key.

* **Localization**
  * Added translatable label for "App ID" in English.

* **Tests**
  * Simplified Gupshup dialog test flow to proceed directly to submission and confirmation assertion.

### DIFF
--- a/src/containers/SettingList/Providers/Providers.test.tsx
+++ b/src/containers/SettingList/Providers/Providers.test.tsx
@@ -276,10 +276,6 @@ describe('gupshup dialog', () => {
       expect(screen.getByText('Loading...')).toBeInTheDocument();
     });
 
-    await waitFor(() => {
-      expect(screen.getByText('Active?')).toBeInTheDocument();
-    });
-
     const saveButton = await screen.findByTestId('submitActionButton');
     await user.click(saveButton);
 


priyanshu6238 (commented):
We are allowing values to be saved without checking whether the active button is enabled, which sets is_active to false in the database.

When a message is sent through the simulator, or when someone tries to preview something, it crashes. This happens because in create_and_send_message, we do:

bsp_credential = Partners.organization(organization_id).services["bsp"]

The crash occurs because, while creating the secrets map, we only include credentials that are marked as active. Since provider ID 1 is set to inactive, the system is unable to fetch the message module via the handler key 

Image

Image

Image

Image

PR #3864, opened 2026-04-01
fix: pin axios to 1.13.2 to prevent upgrade to compromised versions 

priyanshu6238:
## Summary by CodeRabbit

* **Chores**
  * Updated dependency version specification to ensure consistent package resolution.


rvignesh89 (changes requested):
Don't think we need to pin to exact version since the vulnerable packages have been removed from npmjs

priyanshu6238:
Not needed, so closing this now.

PR #3877, opened 2026-04-08
Fix for assistant prompt not updating after knowledge base update  

### LINKED ISSUE #3875: [Bug]: Assistant prompt not updating after knowledge base update
### Summary

When updating an existing assistant, changes to the prompt are not reflected if the knowledge base is updated first. The assistant continues to use the older prompt. However, if the prompt is updated before the knowledge base, it works as expected.

### Steps to Reproduce

1. Open an existing assistant
2. Update the prompt
3. Update the knowledge base
5. Save

### Expected Behaviour

The assistant should use the latest updated prompt, regardless of whether the knowledge base or prompt was updated first.

### Actual Behaviour

The assistant continues to use the older prompt when the knowledge base is updated first and the prompt is updated afterward.

### Acceptance Criteria

Prompt updates should always be reflected correctly

akanshaaa19:
target issue is #3875

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed a duplicate form field assignment to ensure consistent validation.
  * Fixed form reset gating during assistant edits to prevent unintended resets.

* **Tests**
  * Expanded editing tests to cover file management, uploads, and exact notification assertions.

* **Chores**
  * Added and reordered test mocks to support new upload and create flows.


rvignesh89 (approved):
Works as expected but is missing tests

PR #3894, opened 2026-04-10
fix: truncate long assistant names in header 

priyanshu6238:
### Summary 

 The size of the icon should not change based on the size of the assistant name. Assistant name should use fixed width

before:
Screenshot 2026-04-10 at 6 53 43 PM

After :

 
Screenshot 2026-04-10 at 6 54 03 PM

## Release Notes

* **Style**
  * Long assistant names now truncate with ellipsis instead of causing layout overflow
  * Header and breadcrumb elements improved to handle constrained space with proper text truncation
  * Edit buttons and action controls maintain stable positioning in responsive layouts without shifting

### DIFF
--- a/src/containers/Assistants/AssistantDetail/ConfigEditor.module.css
+++ b/src/containers/Assistants/AssistantDetail/ConfigEditor.module.css
@@ -50,7 +50,7 @@
   color: #073f24;
   overflow: hidden;
   text-overflow: ellipsis;
-  white-space: nowrap;
+  white-space: normal;
   min-width: 0;
   flex: 1;
 }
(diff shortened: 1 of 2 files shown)


src/containers/Assistants/AssistantDetail/ConfigEditor.module.css
```diff
@@ -41,18 +41,25 @@
...
+  overflow: hidden;
 }
 
 .Breadcrumb {
   font-size: 1rem;
   font-weight: 600;
   color: #073f24;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
```
rvignesh89:
```suggestion
  white-space: normal;
  word-break: break-word;
```
Don't think truncating is a good idea

rvignesh89 (approved):
Lets not truncate the name/version of the assistant

PR #3903, opened 2026-04-14
ci: remove e2e test concurrency limits

rvignesh89:
We don't need to limit e2e test concurrency since we've temporarily got ngrok license for unlimited sessions so we are not limited by the 3 concurrent endpoint limits anymore.


priyanshu6238 (commented):
If we have the license, can we run 3 parallel jobs instead of 2? It would reduce the total time to around 20 minutes.

priyanshu6238 (approved):

PR #3906, opened 2026-04-15
fix: fetch knowledge base from assistant version instead of assistant

mdshamoon:
## Summary

- Moved `vectorStore` (knowledge base data) from the assistant-level `GET_ASSISTANT` query to the version-level `GET_ASSISTANT_VERSIONS` query, so each version carries its own knowledge base files
- Removed the separate `vectorStore` prop from `ConfigEditor` — it now derives it directly from `version.vectorStore`, eliminating stale closure bugs
- Replaced the post-mount `useEffect` + `resetForm` approach with `computedInitialValues` so `useFormik` starts with correct files synchronously on mount (fixes `KnowledgeBaseOptions` local state initializing with empty files)
- Added `VectorStore: { keyFields: false }` to Apollo `InMemoryCache` type policies to prevent Apollo normalizing all `VectorStore` objects with the same `id` into a single shared cache entry — which was causing all versions to show the same files
- Added `searchParameter={['name_or_assistant_id']}` for assistant list search

## Test plan

- [ ] Open an assistant with multiple versions that have different knowledge base files
- [ ] Switch between versions — each version should show its own knowledge base files and file count
- [ ] Verify creating a new assistant still works
- [ ] Verify saving a version (updating knowledge base) still works
- [ ] Run `npx vitest run src/containers/Assistants/` — all tests pass

## Summary by CodeRabbit

* **New Features**
  * Vector store information, including associated files and metadata, is now retrieved and managed as part of assistant versions.

* **Refactor**
  * Improved how vector store data is derived and initialized during assistant configuration.


src/config/apolloclient.ts
```diff
@@ -23,6 +23,9 @@ import * as Sentry from '@sentry/react';
 
 export const cache = new InMemoryCache({
   typePolicies: {
+    VectorStore: {
+      keyFields: false,
```
mdshamoon, line 27: All vector store have the same id with a different knowledge base id. Unsure why is that.

rvignesh89 (approved):

src/config/apolloclient.ts
priyanshu6238: what we are doing here ?

mdshamoon: when we are fetching the vector store the assistant uses the same vector store but different versipns for the vector store. Apollo caches based on vector store id so it was fetching the same data for assistant versions even if they have different version of the vector store

PR #3908, opened 2026-04-15
fix: use updateAssistantAttrs for assistant rename to prevent new version creation

mdshamoon:
## Summary

- Added `UPDATE_ASSISTANT_ATTRS` mutation which calls the new backend `updateAssistantAttrs` API
- Switched `AssistantDetail` rename flow from `UPDATE_ASSISTANT` to `UPDATE_ASSISTANT_ATTRS` — the old `updateAssistant` mutation was creating a new assistant version on every rename because the backend treats any `updateAssistant` call as a version save
- Updated mocks accordingly

## Test plan

- [ ] Rename an assistant from the detail page — verify no new version is created in the version panel
- [ ] Verify saving actual version changes (instructions, model, temperature) still creates a new version
- [ ] Run `npx vitest run src/containers/Assistants/AssistantDetail/` — all 57 tests pass


rvignesh89 (approved):

mdshamoon:
This is not needed anymore

PR #3925, opened 2026-05-05
Bump heroku-buildpack-nodejs to v344 and tighten node engines

AmishaBisht:
**Description**
 
Prepares the frontend for the gigalixir-24 stack upgrade by:
 - bumping heroku-buildpack-nodejs from v183 → v344
 - tightening [engines.node to >=22.12.0 <23](https://nodejs.org/en/about/previous-releases) (so the buildpack stops picking incompatible Node 22.0.0)
 - pinning all CI workflows to read Node version from .tool-versions()ocal dev (asdf/nvm/mise reads .tool-versions)       and CI now use the exact same Node)
 - yarn.lock changes are cosmetic alias consolidation only- no package versions changed.

**Test plan**

  - Unit + E2E tests pass on CI with Node 22.21.1
  - Staging gigalixir build succeeds on current stack, then on gigalixir-24
  - Staging app smoke test: login, flow editor, chat, GraphQL all work

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,2 +1,2 @@
-https://github.com/heroku/heroku-buildpack-nodejs.git#v244
+https://github.com/heroku/heroku-buildpack-nodejs.git#v344
 https://github.com/heroku/heroku-buildpack-nginx.git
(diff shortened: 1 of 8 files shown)


rvignesh89 (approved):
Let's do a thorough testing in staging. Last time we did this we noticed staging release didn't go through so also verify in Gigalixir UI that the release went through.

PR #3929, opened 2026-05-07
fix: prevent duplicate error notifications on flow publish

### LINKED ISSUE #3499: Flow publish - Same error showing two times
In the Consent flow, we’ve added a Stop flow so that users can opt out if they want. Since the option to type “stop” is already there inside the flow, it is showing an error. 
**The issue is that the same error is showing two times.**

Image

srushti-55:
When clicking "Publish & go back", publishLoading was never set, leaving both publish buttons active. A second click would fire a second mutation, and both onError callbacks would call setNotification, showing the same error toast twice.

Guard handlePublishFlow with publishLoading and centralise setPublishLoading(true) there so no duplicate mutations can fire regardless of which publish button is clicked.

Fixes glific#3499

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate flow publish actions from being initiated while a publish is already in progress, avoiding concurrent publish conflicts and accidental double-submissions.

* **Tests**
  * Added a test that verifies the UI blocks additional publish attempts while the first publish is loading and displays the appropriate error dialog when a second attempt is made.

[Review Change Stack](https://app.coderabbit.ai/change-stack/glific/glific-frontend/pull/3929)


priyanshu6238:
E2E ci test case will fail. you can ignore it

priyanshu6238 (changes requested):
test case is failing can you fix it?

srushti-55:
Hi, could you please approve the CI workflows to run?

priyanshu6238:
@srushti-55 The problem here is that if the error is occurring at two places, the same message is getting shown twice. You just need to filter out the duplicate messages. Try to replicate the error locally you’ll understand it better that way.

priyanshu6238:
If possible, can you also write down the steps you used to verify that it’s working properly in local? It will help while reviewing/testing the fix.

srushti-55:
Thanks for the review! To clarify the root cause: the duplicate notification happens because the original
   state-based guard (if (publishLoading) return) doesn't work when both buttons are clicked synchronously — React
   hasn't re-rendered yet, so publishLoading is still false for both clicks, and two separate mutations fire. Both then
   call onError, showing the same toast twice.
  
   I fixed this with useRef because it's synchronous — the ref is set immediately in the same event tick, so the second
   click is blocked before a second mutation fires. This also prevents an unnecessary duplicate API call to the
   server.
  
   That said, if you'd prefer the deduplication approach (filtering duplicate messages at the notification call site),
   I'm happy to switch to that. Could you confirm which approach you'd like me to go with?
   
   these are the steps
  1. Open any flow in the editor
  2. Click Publish to open the publish dialog
  3. Click "Publish & stay" and immediately click "Publish & go back" before the response returns
  4. Only one error toast appears (previously two appeared)
  5. The network tab shows only one publishFlow mutation request

priyanshu6238:
Screenshot 2026-05-23 at 9 38 12 AM

No, the mutation is not getting triggered twice. The issue here is that every node is being validated, which is why the same error message is appearing multiple times. We just need to remove the duplicate elements from the error list. You can also check the Network tab for confirmation.

I just tested your changes locally, and I am still getting duplicate messages, as you can see in the above image.

priyanshu6238:
@srushti-55  Closing this for now. Feel free to reopen it once all the comments have been resolved.

PR #3951, opened 2026-05-26
fix: update error handling navigation and button labels in ErrorBoundary

### LINKED ISSUE #3938: System crash/logout when contact profile is deleted or unavailable
## Description

The frontend crashes and, in some cases, logs the user out when trying to access a contact profile that no longer exists. This happens because the application attempts to read activeProfile from a null response.

### Sentry Error:
TypeError: Cannot read properties of null (reading 'activeProfile')

This issue was reproduced in production for the Udyham instance where contact profile ID 2 had been deleted.

### Steps to Reproduce

1. Scenario 1 — Deleted contact while page is open
2. Open a contact profile page in the browser.
3. Delete the same contact/profile from the backend/admin panel.
4. Refresh the page or trigger a re-fetch.
5. API returns null, but the frontend still tries to access activeProfile.
6. Instance crashes or user gets logged out.

Scenario 2 — Browser history/bookmarked URL

1. Open a previously saved/bookmarked contact profile URL.
2. Delete that contact/profile from backend/admin.
3. Revisit the URL.
4. Frontend enters a broken state.

Scenario 3 — Race condition

1. User has the contact profile page open.
2. Contact gets deleted during the session.
3. Any re-query/re-render occurs.
4. Frontend crashes because the record no longer exists.

### Expected Behavior

1. Application should gracefully handle missing/deleted contact profiles.
2. Show a proper “Contact not found” or empty state.
3. Redirect user safely if required.
4. User session should remain intact.
5. No frontend crash should occur.

### Actual Behavior

1. Frontend crashes with:
    Cannot read properties of null (reading 'activeProfile')
2. In some cases, the user gets logged out.
3. Possible Root Cause

Frontend assumes the contact profile object is always present and does not handle null responses from the API.

### Suggested Fix

1. Add null/undefined checks before accessing activeProfile.
2. Handle missing profile responses gracefully.
3. Add fallback UI for deleted/non-existent contacts.
4. Prevent session/logout side effects caused by this error.

### Additional Notes

1. Reproduced in production.
2. Related Sentry issue:
3. [GLIFIC-FRONTEND-258](https://project-tech4dev.sentry.io/issues/7478244206/?alert_rule_id=16154124&alert_type=issue&environment=production&notification_uuid=f0719ea5-aab0-4cc5-ab43-dd12838e5e49&project=5296557&referrer=discord)

akanshaaa19:
target issue is #3938 

Instead of adding fixes to the contact module, I replaced the "Logout" button in the ErrorBoundary with a "Close" button that redirects to /chat. Users no longer get logged out when this error surfaces, which is a better UX given how rarely this issue actually occurs.

## Summary by CodeRabbit

* **Improvements**
  * Updated error dialog button behavior: "Close" navigates to chat and dismisses the error, while "Retry" reloads the page.

[Review Change Stack](https://app.coderabbit.ai/change-stack/glific/glific-frontend/pull/3951?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/components/errorboundary/ErrorBoundary.tsx
+++ b/src/components/errorboundary/ErrorBoundary.tsx
@@ -48,13 +48,14 @@ class ErrorBoundary extends Component<any, any> {
                 title="Error !"
                 colorCancel="warning"
                 handleOk={() => {
-                  window.location.reload();
-                }}
-                handleCancel={() => {
+                  this.setState({ hasError: false });
                   navigate('/chat');
                 }}
-                buttonOk="Retry"
-                buttonCancel="Close"
+                handleCancel={() => {
+                  window.location.reload();
+                }}
+                buttonOk="Close"
+                buttonCancel="Retry"
                 alignButtons="center"
                 contentAlign="center"
               >
--- a/src/components/errorboundary/ErrorBoundary.tsx
+++ b/src/components/errorboundary/ErrorBoundary.tsx
@@ -51,10 +51,10 @@ class ErrorBoundary extends Component<any, any> {
                   window.location.reload();
                 }}
                 handleCancel={() => {
-                  navigate('/logout/user');
+                  navigate('/chat');
                 }}
                 buttonOk="Retry"
-                buttonCancel="Logout"
+                buttonCancel="Close"
                 alignButtons="center"
                 contentAlign="center"
               >
--- a/src/components/errorboundary/ErrorBoundary.test.tsx
+++ b/src/components/errorboundary/ErrorBoundary.test.tsx
@@ -9,18 +9,24 @@ vi.mock('react-router', async () => ({
   useNavigate: () => mockedUsedNavigate,
 }));
 
+let shouldThrow = true;
 const ThrowingChild = () => {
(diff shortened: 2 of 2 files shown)


src/components/errorboundary/ErrorBoundary.tsx
```diff
@@ -51,10 +51,10 @@ class ErrorBoundary extends Component<any, any> {
                   window.location.reload();
                 }}
                 handleCancel={() => {
-                  navigate('/logout/user');
+                  navigate('/chat');
```
priyanshu6238: Screenshot 2026-05-26 at 10 03 36 AM

Right now, when I delete the contact and try to access that profile again, I am getting this error. On clicking on retry ,close and even after clicking the cross icon on the dialog box, the dialog is not closing and I keep getting the same error .But after clicking 2-5 times on retry i move to the chat screen

akanshaaa19: this is fixed

priyanshu6238 (approved):

PR #3988, opened 2026-06-19
fix: Flow publish - Same error showing multiple times

### LINKED ISSUE #3499: Flow publish - Same error showing two times

nthapa000:
### Problem

When a flow uses a keyword (e.g., `stop`) that is already configured as a keyword for another flow (e.g., the Stop/Consent flow), publishing the flow shows a validation error:

> "stop" has already been used as a keyword for a flow

However, if `stop` appears in **multiple places** within the flow — such as in two or more "Wait for Response" nodes or as an option — the **same error message is displayed multiple times**, once for each occurrence.

For example, if `stop` is expected in 3 places, the user sees the same error 3 times, which is redundant and confusing.

### Root Cause

The `publishFlow` mutation returns one error object per occurrence of the conflicting keyword in the flow. The `errorMsg()` function in `FlowEditor.tsx` was rendering **all** errors directly without filtering for uniqueness, so duplicate messages from the backend were rendered as-is.

### Solution

Added a `.filter()` step in the `errorMsg()` function to deduplicate errors by their `message` field before rendering. This ensures each unique validation error is displayed **only once**, regardless of how many times the keyword appears in the flow.

```tsx
const errorMsg = () => (
  <div className={styles.DialogError}>
    {flowValidation
      .filter(
        (msg: any, i: number, self: any[]) =>
          self.findIndex((m: any) => m.message === msg.message) === i
      )
      .map((message: any) => (
        <div key={message.message} className={styles.ErrorMsg}>
          <WarningIcon className={styles.ErrorMsgIcon} />
          {message.message}
        </div>
      ))}
  </div>
);
```

### Changes

| File | Change |
|------|--------|
| `src/components/floweditor/FlowEditor.tsx` | Added `.filter()` to deduplicate validation errors by message before rendering |
| `src/mocks/Flow.tsx` | Added `publishFlowWithDuplicateErrors` mock returning 5 identical error messages |
| `src/components/floweditor/FlowEditor.test.tsx` | Added test to verify that duplicate errors from the backend are rendered only once |

### Test added

**`publish flow with duplicate errors shows each error only once`**

- Mocks the `publishFlow` mutation to return **5 identical** error messages
- Clicks Publish → confirms → waits for error dialog
- Asserts that only **1** error message element is rendered in the DOM

### Test results

All tests in `FlowEditor.test.tsx` pass ✅ (17 passed, 1 skipped).

> **Note:** `ChatInterface.test.tsx` has a pre-existing test failure that is **unrelated** to this PR.

### Before & After

| Before | After |
|--------|-------|
| Same error shown N times (once per occurrence in the flow) | Each unique error shown only once |

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where duplicate publish-flow validation messages could display repeatedly.
  * Publish error and warning messages are now de-duplicated so each identical message appears only once.
* **Tests**
  * Added a test case to ensure the publish error UI correctly de-duplicates repeated validation error messages.

[Screencast from 18-06-26 01:08:24 PM IST.webm](https://github.com/user-attachments/assets/3c5622ba-635b-483e-9963-caad67b42069)


src/components/floweditor/FlowEditor.tsx
```diff
@@ -301,12 +301,14 @@ export const FlowEditor = () => {
...
   const errorMsg = () => (
     <div className={styles.DialogError}>
-      {flowValidation.map((message: any) => (
-        <div key={message.message} className={styles.ErrorMsg}>
-          <WarningIcon className={styles.ErrorMsgIcon} />
-          {message.message}
-        </div>
-      ))}
+      {flowValidation
+        .filter((msg: any, i: number, self: any[]) => self.findIndex((m: any) => m.message === msg.message) === i)
```
priyanshu6238: Can you use a Set or Map here to deduplicate the messages? The current implementation uses findIndex inside filter, which makes it O(n²). It would be better to optimize this to O(n).

nthapa000: fix: https://github.com/glific/glific-frontend/pull/3988/changes/323ec43b1d50c2a401c3050d9a0b50f42b117041

priyanshu6238 (approved):
LGTM

PR #3991, opened 2026-06-21
feat: embed Superset analytics dashboard

rvignesh89:
## Summary

- **Embedded Superset dashboard**: Adds a `/analytics` page that embeds a Superset guest dashboard via `@superset-ui/embedded-sdk`
- **Token refresh**: Proactively refreshes the embed token every 4.5 min via `setInterval`; fixes `fetchGuestToken` to be an async callback (not a static string) so the SDK can call it on token expiry
- **Loading state**: Shows Glific's standard `<Loading />` spinner while the initial token is being fetched; dashboard container stays in the DOM throughout so `getElementById` works during the async effect
- **Graceful error handling**: Renders a themed error page (Glific colors/typography) with "refresh the page" and "contact support" actions when token fetch fails
- **Sentry tracking**: Reports token fetch failures as captured Sentry events via `setLogs(..., 'error', true)`
- **Feature flag**: Data Analytics sidebar item is gated behind `supersetEnabled` org service flag; adds field to `GET_ORGANIZATION_SERVICES` GQL query and `ServiceType` union

## Results
image

## Test plan

- [ ] `npx vitest run src/containers/Analytics/Analytics.test.tsx` — 6 tests pass (loading spinner, embed success, `fetchGuestToken` callback, error UI, Sentry logging, interval cleanup on unmount)
- [ ] With `supersetEnabled: true` in `localStorage.organizationServices` — Data Analytics appears in sidebar for Manager+ roles
- [ ] With `supersetEnabled: false` or missing — Data Analytics does not appear in sidebar
- [ ] On `/analytics` — loading spinner appears briefly, then dashboard iframe renders
- [ ] Temporarily break `ANALYTICS_ENDPOINT` in `.env` — error page with refresh + support link renders (no blank page)
- [ ] Token refresh: observe `POST /v1/get-embed-token` firing every ~4.5 min in DevTools Network tab


.claude/settings.json
```diff
@@ -1,5 +1,12 @@
 {
   "enabledPlugins": {
     "claude-md-management@claude-plugins-official": true
+  },
+  "worktree": {
+    "symlinkDirectories": [
+      "node_modules",
+      "build",
+      ".env"
```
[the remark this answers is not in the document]
shijithkjayan: @rvignesh89

.cursorignore
```diff
@@ -0,0 +1 @@
+.env
```
shijithkjayan, line 1: Do we have to push this to the repo?

shijithkjayan (approved):

.cursorignore
rvignesh89: Yep, otherwise cursor is able to read these files

PR #4006, opened 2026-06-25
feat: add ConfigurePageStub for flow configuration and update routing

### LINKED ISSUE #5286: Page loads Indefinitely with undefined flow UUID after editing a flow from the flow creation page and then hitting configure
When a user clicks on the `Edit` option for a flow and makes changes on the flow creation page, clicking `Configure` directly afterward causes the page to keep loading. The URL path ends with `undefined` instead of the flow's UUID. However, upon clicking the browser's Back button, the changes are found to have been saved successfully.

Steps to Reproduce:

>> Choose an existing flow.
>> Select the `Edit` option.
>> Click on `Configure`.

Observed Behavior:
The page continues loading indefinitely, and the URL ends with `undefined` instead of the flow UUID.

priyanshu6238:
## Summary 

Closes: https://github.com/glific/glific/issues/5286

* **Bug Fixes**
  * Improved flow editing behavior so configured links update correctly after selecting a linked item (including when updating, not just copying).
  * After editing a flow, the app now reliably navigates to the configure screen and displays the expected flow identifier.

* **Tests**
  * Expanded test coverage for the new configure-flow route and added assertions for the expected post-edit navigation and on-screen identifier.

## Summary by CodeRabbit

* **New Features**
  * Added support for navigating to a flow configuration page after editing a flow.

* **Bug Fixes**
  * Improved update handling so the correct link is retained for all saved items, not just copy actions.

### DIFF
--- a/src/containers/Form/FormLayout.tsx
+++ b/src/containers/Form/FormLayout.tsx
@@ -43,7 +43,6 @@ export interface FormLayoutProps {
   additionalAction?: any;
   additionalQuery?: Function | null;
   linkParameter?: any;
-  setLinkOnUpdate?: boolean;
   cancelLink?: string | null;
   languageSupport?: boolean;
   setPayload?: Function;
@@ -116,7 +115,6 @@ export const FormLayout = ({
   additionalQuery = null,
   defaultAttribute = null,
   additionalAction,
-  setLinkOnUpdate = false,
   icon,
   idType = 'id',
   additionalState,
@@ -240,7 +238,7 @@ export const FormLayout = ({
         customError.setErrors(codeErrors);
       }
     } else {
-      if ((type === 'copy' || setLinkOnUpdate) && updatedItem) setLink(updatedItem[linkParameter]);
+      setLink(updatedItem[linkParameter]);
       if (additionalQuery) {
         additionalQuery(itemId);
       }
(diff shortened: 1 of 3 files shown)


src/containers/Form/FormLayout.tsx
```diff
@@ -238,7 +240,7 @@ export const FormLayout = ({
         customError.setErrors(codeErrors);
       }
     } else {
-      if (type === 'copy') setLink(updatedItem[linkParameter]);
+      if ((type === 'copy' || setLinkOnUpdate) && updatedItem) setLink(updatedItem[linkParameter]);
```
akanshaaa19: we can just remove this condition, since link is not used anywhere else- its only used for redirecting. we wont need another param here
 
```suggestion
      setLink(updatedItem[linkParameter]);
```

akanshaaa19: can you check what happens if we remove it

akanshaaa19 (changes requested):

src/containers/Form/FormLayout.tsx
priyanshu6238: Yes, I added console logs and checked all the pages that use the type prop in the Layout component. It doesn't seem to be used anywhere, so it should be safe to remove.
fixed: https://github.com/glific/glific-frontend/pull/4006/commits/89a18c09f9afbac87c88edfc816681f4f477a36e

akanshaaa19 (approved):

PR #4007, opened 2026-06-25
chore: tighten Content Security Policy (report-only rollout)

### LINKED ISSUE #5120: [security]: Tighten Content Security Policy to remove wildcard directives
## Description

The Content Security Policy (CSP) for the Glific frontend is overly permissive. The dev server (configured in `vite.config.ts`) sets `default-src * data:` and `connect-src *` (both wildcards), and includes `'unsafe-eval'` in `script-src`. The production CSP, served via nginx config, should be audited and tightened in the same pass.

This creates two distinct attack surfaces:
1. Any injected script can load resources from or exfiltrate data to arbitrary external domains.
2. `unsafe-eval` permits dynamic code execution (e.g., via `eval()`, `Function()`), which is exploitable in XSS attacks.

The CSP must be tightened in both the dev server config and the nginx config to use explicit allowlists for all known integrations while removing wildcard permissiveness. Known integrations that must remain functional include: PostHog, Sentry, AppSignal, Stripe, Google Fonts / reCAPTCHA, and Canva embeds.

## Acceptance Criteria

- [ ] `default-src` no longer uses a wildcard (`*`); it falls back to `'self'` or an explicit allowlist.
- [ ] `connect-src` no longer uses a wildcard; it explicitly lists domains for PostHog, Sentry, AppSignal, Stripe, and the Glific backend WebSocket/HTTP endpoints.
- [ ] `unsafe-eval` is removed from `script-src`; the app is verified to function without it (or a justified, minimal exception is documented).
- [ ] `unsafe-inline` in `script-src` is replaced with a nonce or hash-based approach where feasible, or retained with justification if React's bundling model requires it.
- [ ] Both the dev server CSP (`vite.config.ts`) and the production nginx CSP config are updated and kept in sync.
- [ ] The following integrations are verified to work under the new policy: PostHog, Sentry, AppSignal, Stripe payment flows, Google Fonts, Google reCAPTCHA, and Canva iframe embeds.
- [ ] Before deploying, CSP violation reports (via Sentry or PostHog) are reviewed to identify any domains loaded by org admin browser extensions or other unlisted sources, so legitimate domains are not silently broken by the tighter policy.
- [ ] No browser CSP console errors appear during normal app usage (login, chat, flow editor, billing, onboarding).

## Edge cases

- PostHog host is configured via `VITE_PUBLIC_POSTHOG_HOST` env var and may differ between deployments (e.g., `us.posthog.com` vs `eu.posthog.com`); the nginx CSP must accommodate both or use a templated config per environment.
- Sentry DSN ingest domain varies by Sentry plan/region; the allowlist should cover `*.sentry.io` and `*.ingest.sentry.io`.
- The flow editor may load additional third-party resources (e.g., CDN assets); these must be audited and added if needed.

## Open Questions

- Should CSP violation reporting be enabled via a `report-uri` or `report-to` directive, and if so, which endpoint should receive reports?
- Is `unsafe-eval` actually required by any bundled dependency (e.g., the flow editor), or is it a leftover from an earlier setup? This must be audited before removal.

shijithkjayan:
Addresses glific/glific#5120.

## Summary

The CSP previously used wildcards (`default-src * data:`, `connect-src *`), so any injected script could load from / exfiltrate to arbitrary domains. This PR replaces them with an explicit allowlist.

To avoid breaking production for any NGO, the tightened policy is rolled out in **report-only** mode first: the current permissive policy stays **enforced**, and the strict policy ships as `Content-Security-Policy-Report-Only`. The browser reports what it *would* block without blocking anything. Once reports are clean, a small follow-up promotes it to enforced.

Both the dev (`vite.config.ts`) and prod (`config/nginx.conf.erb`) policies are updated and kept in sync.

## What the strict policy does

- `default-src 'self'` and an explicit `connect-src` (no wildcards): backend (`*.glific.com` + `wss://`), PostHog, Sentry, Stripe, Google, moonshine, CORS proxy.
- `img-src` / `media-src` intentionally stay broad (`https: data: blob:`) — chat renders WhatsApp media hosted on arbitrary external CDNs. (Broad image loading is low-risk; broad *script* loading is not, and scripts are tightened.)
- `unsafe-eval` retained with documented justification: the flow editor's `@nyaruka/temba-components` evaluates event handlers via `new Function(...)`, which has no nonce/hash alternative. Verified its removal breaks the flow editor.
- `unsafe-inline` retained (Vite runtime + MUI/emotion inject inline scripts/styles).
- Added `object-src 'none'` and `base-uri 'self'` hardening.

## Deployment

Set **one** env var per environment to start the observation phase:

- `CSP_REPORT_URI` — where violation reports are POSTed (recommended: a Sentry CSP endpoint, `https://oXXX.ingest.sentry.io/api/<project>/security/?sentry_key=<key>`). If unset, **no report-only header is emitted and this PR is a no-op in production.**

Backend domains are hardcoded (`*.glific.com`) since all deployments live under `*.glific.com` — no custom domains.

## Follow-up (not in this PR — the report-only phase by design)

1. Set `CSP_REPORT_URI` on staging/prod.
2. Let it bake; review CSP reports in Sentry across login / chat / flow editor / billing / onboarding.
3. Add any legitimate domain that gets flagged to the allowlist.
4. **Promote to enforced**: swap the strict policy into the `Content-Security-Policy` header and delete the permissive policy + report-only block. This step closes the "no wildcards" acceptance criteria in glific/glific#5120.

## Testing

- Verified `nginx.conf.erb` renders correctly both with and without `CSP_REPORT_URI`.
- Statically confirmed `unsafe-eval` is required by the bundled flow editor (`new Function` in `@nyaruka/temba-components`).
- Local click-through (login/chat/flow editor) — report-only surfaces violations as console warnings in dev.
- **Report ingestion verified end-to-end**: with `VITE_CSP_REPORT_URI` pointed at the Sentry Security Header endpoint, a **deliberately induced** `frame-src` violation was captured in Sentry — project `GLIFIC-FRONTEND-2EP`, `development` environment, filtered by `event.type:csp`, showing the issue **"Blocked 'frame-src' from 'www.youtube.com'"** (1 event / 1 user). This confirms the browser → `report-uri`/`report-to` → Sentry path works and the same wiring will collect reports in staging/prod via `CSP_REPORT_URI`.

Sentry CSP report ingestion — Blocked 'frame-src' from www.youtube.com

> **Note:** this YouTube `frame-src` violation was **self-inflicted to test the reporting pipeline**, not a real gap — `https://www.youtube.com` is already in the allowlist. It only surfaced because the running dev server was serving a pre-fix policy; it's here purely as proof that report ingestion into Sentry works.

## Summary by CodeRabbit

* **New Features**
  * Added a tighter report-only Content Security Policy for supported environments to help surface potential browser security issues early.
  * Enabled CSP violation reporting so problems can be monitored without changing the currently enforced policy.
  * Updated the development server to apply the same report-only behavior for local testing and validation.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,3 +10,9 @@ VITE_PUBLIC_POSTHOG_PROJECT_TOKEN='Your posthog project token'
 VITE_PUBLIC_POSTHOG_HOST='https://us.i.posthog.com'
 VITE_AI_EVALS_INTRO_VIDEO_URL=''
 VITE_SUPERSET_DASHBOARD_ID='superset dashboardId'
+# Optional: where the dev server POSTs CSP violation reports (Content-Security-Policy-Report-Only).
+# Leave empty to only log violations to the browser console. Set it to a Sentry Security Header
+# endpoint derived from your Sentry DSN to test CSP report ingestion locally:
+#   DSN      https://<KEY>@o<ORG>.ingest.sentry.io/<PROJECT>
+#   endpoint https://o<ORG>.ingest.sentry.io/api/<PROJECT>/security/?sentry_key=<KEY>&sentry_environment=local
+VITE_CSP_REPORT_URI=''
--- a/config/nginx.conf.erb
+++ b/config/nginx.conf.erb
@@ -28,7 +28,45 @@ http {
     add_header X-XSS-Protection "1; mode=block" always;
     add_header X-Frame-Options "deny" always;
     add_header Strict-Transport-Security "max-age=63072000; includeSubdomains; preload" always;
+<%-
+      # --- Content Security Policy ----------------------------------------------------------
+      # Rollout is in two stages so production behaviour cannot break:
+      #   1. The current PERMISSIVE policy is still ENFORCED (the Content-Security-Policy
+      #      header below) — production behaves exactly as before.
+      #   2. The TIGHTENED policy is sent as Content-Security-Policy-Report-Only: the browser
+      #      does NOT block on it, it only reports what it *would* block to CSP_REPORT_URI
+      #      (recommended: a Sentry CSP endpoint). When those reports are clean, replace the
+      #      enforced policy below with `strict_csp` and drop this report-only block.
+      # Keep `strict_csp` in sync with the dev-server CSP in vite.config.ts.
+      #
+      # The backend is served per deployment at api.<host> under *.glific.com (no custom
+      # domains), so `https://*.glific.com wss://*.glific.com` covers every backend HTTP +
+      # WebSocket origin. CSP wildcards are only valid as the leftmost label, and
+      # `*.glific.com` matches multi-level hosts like api.tides.glific.com.
+      #
+      # CSP_REPORT_URI: where violation reports are sent; if unset, no report-only header is
+      #   emitted and this change is a no-op in production.
+      # `unsafe-eval`: required by the flow editor (@nyaruka/temba-components uses
+      #   `new Function(...)` for event handlers); there is no nonce/hash alternative.
+      csp_report_uri = ENV['CSP_REPORT_URI']
+      strict_csp =
+        "default-src 'self'; " \
+        "script-src 'self' 'unsafe-inline' 'unsafe-eval' blob:; " \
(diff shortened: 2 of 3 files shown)


rvignesh89 (approved):
I don't remember how the report only header works, but I suggest deploy this to staging once and verify that none of the features break. Let's also hold off this release until end of next week

PR #4018, opened 2026-06-30
fix: enhance Sentry error handling and configuration

### LINKED ISSUE #4020: Clean up sentry errors
we are getting lots of errors from sentry. some of them are not even needed- so we need to clean up a little bit and only have important errors come to discord

akanshaaa19:
## Why

~99% of our Sentry volume (**166k events / 14 days**) was non-actionable: console
output and React dev warnings shipped as errors. This PR removes that noise at the
source, improves error grouping, and fixes a chat subscription crash surfaced in
the noise.

### 1. Sentry noise reduction — `src/sentry.config.ts`
- **Removed `captureConsoleIntegration` entirely** — every React dev warning is a
  console.error, so this captured them all. Genuine errors still arrive via thrown
  exceptions (global handlers + <Sentry.ErrorBoundary>) and explicit captureException().
- **`ignoreErrors`** drops known-benign noise: ResizeObserver loops and opaque
  cross-origin `"Script error."` events.
- **`denyUrls`** drops errors from browser extensions / injected scripts.
- **No `beforeSend`** — we intentionally do NOT filter by level (non-error events
  come only from deliberate `setLogs(..., true)` calls) and do NOT drop stack-less
  events (that would also drop `httpClientIntegration` HTTP failures).
- **`tracesSampleRate` → `0.5` in production** (was `1`); staging/dev stay at `1.0`.

### 2. GraphQL error grouping — `src/config/apolloclient.ts`
- `captureException` now sends a real `Error` with a `fingerprint` on the message,
  so one error stops fragmenting into a separate issue per `path`/`location`.

## Expected impact

| | Before | After (expected) |
|---|---|---|
| Console / React-warning events | ~99% of 166k / 14d | **0** (no console capture) |
| Non-error-level events | shipped | dropped in `beforeSend` |
| Cross-origin `"Script error."` | shipped | dropped (no stack) |
| Extension-origin errors | shipped | dropped (`denyUrls`) |
| GraphQL errors | 1 issue per path/location | grouped by message |

Net: from 166k/14d down to roughly the genuine-error tail (low thousands), with
real exceptions and explicit `captureException` calls preserved.

## Behaviour to be aware of
- `console.error(someError)` calls that aren't re-thrown no longer reach Sentry
  (e.g. token-refresh `handleError`). Important paths use explicit
  `captureException`, so this is intentional.
- Extension filtering relies on `denyUrls` **and** the built-in "browser extensions"
  Inbound Filter — confirm that filter is enabled in project settings.

## Testing
- `npx tsc --noEmit` — clean
- Full suite: **1390 passed, 12 skipped, 0 failures** (`npx vitest run`)
- (`yarn test:no-watch` is blocked locally by a Node-version engine check; run via
  `npx vitest` or Node 22.)

## Notes
- Replay sample rates remain set but inert — no `replayIntegration()` is registered,
  so no replays are recorded (unchanged from before).
- Do not auto-merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Sentry error reporting with more structured context (message, locations, path) and better issue fingerprinting.
  * Added filtering to drop low-signal, stackless “Script error” style events and ignore known benign errors.
  * Refined logging so non-string values are rendered more reliably, and captured events include consistent breadcrumbs.
  * Reduced unnecessary informational auth-status logging when tokens are still valid.


src/sentry.config.ts
```diff
@@ -1,20 +1,51 @@
 import * as Sentry from '@sentry/react';
 
+const environment = import.meta.env.VITE_ENV || 'development';
+const isProduction = environment === 'production';
+
+const KNOWN_BENIGN_EXCEPTIONS: (string | RegExp)[] = [
+  'ResizeObserver loop limit exceeded',
+  'ResizeObserver loop completed with undelivered notifications',
```
shijithkjayan, line 8: What does these exceptions mean?

src/sentry.config.ts
```diff
@@ -1,20 +1,51 @@
...
     Sentry.browserSessionIntegration(),
     Sentry.httpClientIntegration(),
-    Sentry.captureConsoleIntegration(),
   ],
+
+  ignoreErrors: KNOWN_BENIGN_EXCEPTIONS,
+  denyUrls: DENY_URLS,
+
+  beforeSend(event) {
+    if (event.level && !['error', 'fatal'].includes(event.level)) return null;
```
priyanshu6238: Someplace  in the codebase we are calling  
```
setLogs(\validateURL failed: ${error}`, 'warning', true);`
````

In logs.ts:16, Sentry.captureMessage(logMessage, type) passes type as the second argument, which Sentry uses to set event.level. In this case, the level is set to 'warning'. it wil going to drop  all this message

File: src/sentry.config.ts [hunk 3]
```diff
@@ -1,20 +1,51 @@
...
+
 Sentry.init({
   dsn: import.meta.env.VITE_SENTRY_DSN,
-  environment: import.meta.env.VITE_ENV || 'development',
-  // Setting this option to true will print useful information to the console while you're setting up Sentry.
+  environment,
   debug: false,
 
-  tracesSampleRate: 1,
+  tracesSampleRate: isProduction ? 0.1 : 1,
```
priyanshu6238: Can we increase it to 50–100%? Otherwise, in production we're tracing only 10% of the requests.

priyanshu6238: Do we want to trace everything on staging ?

akanshaaa19: Bumped to 0.5 (50%) in production.

akanshaaa19: Kept staging/dev at 1.0; traffic there is low. its good for pre-prod perf checks.

src/sentry.config.ts
akanshaaa19: They're benign browser warnings, not app errors. Fired when a ResizeObserver callback itself changes layout, so the next resize can't be delivered in the same frame, a known false positive from UI libs with zero user impact. Sentry's docs recommend ignoring them.

src/sentry.config.ts
akanshaaa19: yess, fixed this

src/sentry.config.ts
```diff
@@ -1,20 +1,51 @@
...
     Sentry.browserSessionIntegration(),
     Sentry.httpClientIntegration(),
-    Sentry.captureConsoleIntegration(),
   ],
+
+  ignoreErrors: KNOWN_BENIGN_EXCEPTIONS,
+  denyUrls: DENY_URLS,
+
+  beforeSend(event) {
+    if (event.exception && !hasUsableStack(event)) return null;
```
shijithkjayan: `beforeSend` only checks `event.exception && !hasUsableStack(event)` — it doesn't filter by `level` at all. The PR description says "drop any event whose `level` isn't `error`/`fatal`", but that check isn't implemented here.

Concretely, `setLogs(..., 'info', true)` in `ContactHistory.tsx` (lines 34, 49) and `setLogs(..., 'warning', true)` in `AddAttachment.tsx` (line 73) call `Sentry.captureMessage`, which has no `event.exception` — so this guard never fires for them and they ship to Sentry exactly as before. Worth adding an explicit level check here, e.g. `if (event.level && !['error', 'fatal'].includes(event.level)) return null;`, if the intent is really to drop non-error events.

src/sentry.config.ts
```diff
@@ -1,20 +1,51 @@
...
+
 Sentry.init({
   dsn: import.meta.env.VITE_SENTRY_DSN,
-  environment: import.meta.env.VITE_ENV || 'development',
-  // Setting this option to true will print useful information to the console while you're setting up Sentry.
+  environment,
   debug: false,
 
-  tracesSampleRate: 1,
+  tracesSampleRate: isProduction ? 0.5 : 1,
```
shijithkjayan, line 27: PR description says `tracesSampleRate` goes to `0.1` in production, but this sets it to `0.5`. Could be a stale description or an intentional change since drafting — worth confirming which value is actually intended, since it's a 5x difference in expected trace volume/cost versus what's documented.

akanshaaa19: yes the description is stale. 0.5 is intentional. Fixed the description.

src/sentry.config.ts
akanshaaa19: the description was stale. We deliberately removed the level filter after it was pointed out that it would drop intentional setLogs(..., 'warning'/'info', true) telemetry. Updated the description to match.

src/sentry.config.ts
```diff
@@ -1,20 +1,39 @@
 import * as Sentry from '@sentry/react';
 
+const environment = import.meta.env.VITE_ENV || 'development';
+const isProduction = environment === 'production';
+
+const KNOWN_BENIGN_EXCEPTIONS: (string | RegExp)[] = [
+  'ResizeObserver loop limit exceeded',
+  'ResizeObserver loop completed with undelivered notifications',
+  'Script error.',
```
shijithkjayan: This entry is redundant — `eventFiltersIntegration` (formerly `inboundFiltersIntegration`) is part of `getDefaultIntegrations()` and is included automatically here since this `Sentry.init()` call never sets `defaultIntegrations`. Its built-in `DEFAULT_IGNORE_ERRORS` already contains `/^Script error\.?$/` (see `@sentry/core/build/cjs/integrations/eventFilters.js:13`), so this line never changes behavior — Sentry was already dropping `"Script error."` events by default. Harmless, but safe to drop unless it's meant as an explicit belt-and-suspenders note in case defaults change.

shijithkjayan (approved):

src/sentry.config.ts
akanshaaa19: yes, removed

PR #4035, opened 2026-07-03
feat: implement WhatsApp phone management with reconnect functionality and status synchronization

### LINKED ISSUE #5104: Phase 7: Number status dashboard + disconnect/suspended alerting
**Parent:** #5092
**Implementation doc:** [`docs/specs/multi-phone-wa-groups.md`](docs/specs/multi-phone-wa-groups.md)
**Priority:** P1

## Goal

Give NGO staff a Glific-internal dashboard showing the health of every Maytapi-linked phone for their org, and alert them when a phone goes disconnected (Maytapi-side) or suspended (Meta-side).

Today, `wa_managed_phones.status` already exists (text). The `Maytapi.ApiClient.list_wa_managed_phones/1` call exposes Maytapi's status field via `GET /{product_id}/listPhones`. What's missing: (a) we don't poll it on a schedule, (b) there's no UI to surface it, (c) no alerting when it flips to bad.

## Investigation (do this first)

- [x] Read the Maytapi `listPhones` response schema. Confirm the exact set of status values (`active`, `loading`, `qr-screen`, `disconnected`, `expired`, etc.) and which ones correspond to "disconnected at Maytapi" vs "suspended by Meta". Document the mapping in this ticket before coding.
- [ ] Check whether Maytapi delivers a status-change webhook event (e.g. on `ack`, on disconnect). If yes, we can react in real-time; if no, we'll poll. Document.
- [ ] Decide poll interval if needed (suggest: 5 min via Oban Cron, matching other reconciliation jobs).

Mark this ticket as feasibility-confirmed before writing code; if Maytapi can't surface "suspended by Meta" distinctly, document the limitation and proceed with the best signal available.

## Scope

Assuming investigation confirms feasibility:

1. **Status reconciliation job** — Oban Cron worker that calls `list_wa_managed_phones/1` per org, diffs against `wa_managed_phones.status`, updates rows, and fires `Glific.Notifications.create_notification/1` on transitions into bad states.
2. **GraphQL** — `WAManagedPhone.status` (already exists as a column) surfaced on the existing managed-phones query. Add `WAManagedPhone.lastStatusCheckedAt` if helpful.
3. **Frontend** (separate ticket on `glific-frontend`) — managed phones list view with status badges; banner alert on Settings page when any phone is unhealthy.
4. **Notifications** — category `"WA Phone"`, severity `warning` (disconnected) / `critical` (suspended).

## Out of scope

- Auto-recovery / reconnect flow — separate ticket (#TBD reconnect ticket).
- Historical status timeline / audit log.

## Acceptance criteria

- [ ] Investigation section above is filled in with concrete Maytapi status values and webhook availability.
- [x] Status reconciliation job runs on a schedule and updates `wa_managed_phones.status`.
- [x] Transitions into bad states create `Notification` rows with category `"WA Phone"`.
- [x] GraphQL exposes status on existing managed phones queries (no breaking changes).
- [x] Tests cover: status change triggers notification, no-op when status unchanged, multiple phones per org handled independently.
- [x] Frontend peer ticket filed and linked.

## Files likely touched

- `lib/glific/providers/maytapi/api_client.ex` (response parsing)
- `lib/glific/wa_group/wa_managed_phone.ex` (possibly add `last_status_checked_at`)
- New: `lib/glific/jobs/wa_phone_status_worker.ex` (Oban Cron job)
- `lib/glific_web/schema/wa_managed_phone_types.ex` (if status not yet exposed)

## Size

M

## Notes

This is P1 in the broader Maytapi roadmap. Filing under the multi-phone epic for sequencing (Week 6, after the multi-phone failover work in Phase 4 has been running).

### LINKED ISSUE #5105: Phase 7: Reconnect Maytapi phone from Glific (no Maytapi console required)
**Parent:** #5092
**Implementation doc:** [`docs/specs/multi-phone-wa-groups.md`](docs/specs/multi-phone-wa-groups.md)
**Priority:** P1

## Goal

When a Maytapi phone disconnects, let an NGO admin re-link it from Glific without needing to log in to the Maytapi web console.

The blocker today: re-linking requires scanning a fresh QR code from WhatsApp, and Glific has no surface for that.

## Investigation (do this first)

- [ ] Find the Maytapi endpoint that returns the QR/login screen for a given phone — typically named `screen`, `qr`, or `login`. Confirm method, path, response format (image bytes, base64, JSON with QR data URL, etc.).
- [ ] Confirm the lifecycle: does Maytapi automatically issue a new screen when a phone disconnects, or do we need to call a "logout + restart" endpoint first?
- [ ] Check whether the QR has a TTL we need to handle (likely yes — auto-refresh every ~30s on the frontend).
- [ ] Determine if there are auth / permission constraints — only the Maytapi product owner may be able to trigger this; verify with our token.

If the endpoint exists, this is straightforward. If Maytapi requires their console for QR display, this ticket becomes "negotiate with Maytapi for an API or build a thin proxy via headless browser" — flag that as out-of-scope and close.

## Scope

Assuming the endpoint exists:

1. **Backend** — new `Glific.Providers.Maytapi.ApiClient.fetch_phone_screen/2` wrapping the screen/QR endpoint. New GraphQL query `whatsappPhoneScreen(waManagedPhoneId)` returning the QR data (image URL or base64) with `expiresAt`.
2. **Frontend** (separate ticket on `glific-frontend`) — "Reconnect" button on the unhealthy phone row that opens a modal with the QR; auto-polls/refreshes every 30s; closes when status flips to `active` (driven by the Phase 6 status reconciliation job from sibling ticket).
3. **Authorization** — admin only via `Authorize` middleware.

## Out of scope

- Automated reconnect with no user action (not possible — WhatsApp requires scanning a fresh QR).
- Reconnecting via SMS-based pairing (different Maytapi flow, defer).

## Acceptance criteria

- [ ] Investigation findings documented in this ticket.
- [x] `ApiClient.fetch_phone_screen/2` implemented and tested.
- [x] GraphQL query returns QR payload + expiry.
- [x] Admin-only authorization enforced.
- [ ] Tests cover: successful fetch, error response, unauthorized caller.
- [x] Frontend peer ticket filed and linked.
- [x] Manual verification on staging: disconnect a phone in Maytapi, hit Reconnect in Glific, scan QR, watch phone return to `active`.

## Files likely touched

- `lib/glific/providers/maytapi/api_client.ex`
- `lib/glific_web/schema/wa_managed_phone_types.ex`
- `lib/glific_web/resolvers/wa_managed_phone.ex`

## Size

S–M (depends on Maytapi endpoint specifics)

## Notes

Pairs with the status dashboard ticket — that ticket detects the disconnect, this one helps the admin fix it.

akanshaaa19:
target issue is [5104](https://github.com/glific/glific/issues/5104) and [5105](https://github.com/glific/glific/issues/5105)

depends on https://github.com/glific/glific/pull/5333


src/containers/WaGroups/PhoneManagement/PhoneManagement.tsx
```diff
@@ -0,0 +1,158 @@
...
+  label?: string | null;
+  status?: string | null;
+  lastStatusCheckedAt?: string | null;
+}
+
+export const PhoneManagement = () => {
+  const { t } = useTranslation();
+  const roles = getUserRole();
+  const isAdmin = roles.includes('Admin') || roles.includes('Glific_admin');
+  const canSync = isAdmin || roles.includes('Manager');
```
shijithkjayan: **Custom "Dynamic" org roles get menu/route access but the page is functionally dead for them.**

`isAdmin`/`canSync` are computed via literal role-string checks (`'Admin'`, `'Glific_admin'`, `'Manager'`). But the new menu entry (`src/config/menu.ts`, `roles: managerLevel`) includes `'Dynamic'`, and `checkDynamicRole()` in `AuthenticatedRoute.tsx` grants full `adminRoutes` access (including `/group/phones`) to any custom org role. `getUserRole()` never returns the literal string `"Dynamic"` — it returns the org's actual custom role label — so a manager-equivalent custom-role user can navigate here but will see neither the Sync button nor any Reconnect action, with no indication why. Worth reusing/extending a centralized role-permission check (e.g. `getUserRolePermissions()` in `context/role.ts`) instead of raw string matching.

src/containers/WaGroups/PhoneManagement/PhoneManagement.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { GET_WA_MANAGED_PHONES, GET_WA_MANAGED_PHONES_COUNT } from 'graphql/queries/WaGroups';
+import { SYNC_WA_MANAGED_PHONE_STATUSES } from 'graphql/mutations/Group';
+import { SHORT_DATE_TIME_FORMAT } from 'common/constants';
+
+import { ReconnectDialog } from './ReconnectDialog';
+import styles from './PhoneManagement.module.css';
+
+// A phone in these Maytapi states can still send/receive; anything else is
+// surfaced as unhealthy (mirrors the backend classification).
+const HEALTHY_STATUSES = ['active', 'loading'];
```
shijithkjayan: **`HEALTHY_STATUSES` contradicts how two other components classify the same `status` field.**

`StatusBar.tsx` and `PhonesPanel.tsx` both treat only `status === 'active'` as healthy (a `'loading'` phone renders as inactive/red there). Here, `'loading'` is classified as healthy (green badge). The same phone in the same state will show contradictory health signals across three pages of the app. Since the PR comment says this "mirrors the backend classification," it'd be worth having the backend expose a single health flag all three components consume, rather than three independent frontend copies of this classification.

src/containers/WaGroups/PhoneManagement/ReconnectDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+  useEffect(() => {
+    if (!status) return;
+
+    if (status !== 'active') {
+      seenPending.current = true;
+    } else if (seenPending.current) {
+      setNotification(t('WhatsApp phone reconnected.'), 'success');
+      onReconnected();
+    }
+  }, [status]);
```
shijithkjayan: **`seenPending` is trivially satisfied on the very first poll, so this can report "reconnected" and auto-close without any actual reconnect action.**

The Reconnect icon is only shown when `phone.status !== 'active'` (`PhoneManagement.tsx:119`), so this dialog can only ever open on a non-active phone — meaning `seenPending.current = true` gets set by the very first status poll, regardless of whether the admin clicked "Log out & refresh QR." If the phone happens to come back online on its own within the next 20s poll (transient blip, fixed elsewhere, race with another admin), this effect fires a false "WhatsApp phone reconnected" success and auto-closes the dialog. Consider tracking the actual reconnect-mutation call (e.g. a ref set inside `handleReconnect`) rather than inferring intent from the first observed non-active status.

src/containers/WaGroups/PhoneManagement/PhoneManagement.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const syncing = useRef(false);
+
+  const handleSync = async () => {
+    if (syncing.current) return;
+    syncing.current = true;
+    try {
+      const { data } = await syncStatuses();
+      const result = data?.syncWaManagedPhoneStatuses;
+      const errors = result?.errors;
+      if (errors?.length) {
```
shijithkjayan: **A `null` mutation payload is silently treated as success.**

`const errors = result?.errors; if (errors?.length) {...}` — if the backend ever returns `syncWaManagedPhoneStatuses: null` instead of an object with a populated `errors` array (e.g. phone not found, resolver short-circuits), `errors?.length` is `undefined` (falsy), so this falls straight into the success branch and shows a false success toast. Same pattern in `ReconnectDialog.tsx:55-56`. Worth guarding explicitly on `result` being present before treating a missing-errors case as success.

src/containers/WaGroups/PhoneManagement/ReconnectDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+          'warning'
+        );
+        return;
+      }
+      setNotification(t('Logged the phone out. Scan the new QR code to reconnect.'), 'success');
+      refetch();
+    } catch (error) {
+      setErrorMessage(error as Error);
+    }
+  };
```
shijithkjayan, line 70: **`DialogBox`'s Cancel/X/backdrop/Escape aren't gated on this in-flight mutation.**

`buttonOkLoading={reconnecting}` (line 79) only disables the OK button. `handleCancel={onClose}` (line 77) is wired unconditionally to the Cancel button, the X icon, and MUI `Dialog`'s `onClose` (backdrop click + Escape), none of which check `reconnecting`. If the admin clicks "Log out & refresh QR" then immediately closes the dialog before the mutation resolves, the phone gets logged out server-side while the dialog unmounts, and the pending `refetch()`/success toast can fire after unmount — confusing, since the admin no longer has the dialog open to see that a rescan is now needed.

src/containers/WaGroups/PhoneManagement/PhoneManagement.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+const getText = (text: string) => <div className={styles.TableText}>{text}</div>;
+
+const getStatusBadge = (status?: string | null) => (
+  <span className={`${styles.Badge} ${isHealthy(status) ? styles.Healthy : styles.Unhealthy}`}>
+    {status || 'unknown'}
+  </span>
+);
+
+const columnStyles = [styles.Phone, styles.Label, styles.Status, styles.LastChecked];
```
shijithkjayan: **`columnStyles` has 4 entries but there are 5 columns** (`columnNames` below includes `Actions`). Other `List`-based pages with a custom action column (e.g. `WaPollsList.tsx`) define a style for it too. Here the Actions/Reconnect-icon column will get an `undefined` CSS class from `Pager.tsx`'s indexed style lookup — cosmetic, but likely unintended (missing width/alignment for that column).

src/containers/WaGroups/PhoneManagement/PhoneManagement.tsx
```diff
@@ -0,0 +1,158 @@
...
+          errors
+            .map((error: { message?: string }) => error?.message)
+            .filter(Boolean)
+            .join('; '),
+          'warning'
+        );
+        return;
+      }
+      setNotification(result?.message || t('WhatsApp phone statuses have been refreshed.'), 'success');
+      setRefreshList((previous) => !previous);
```
shijithkjayan, line 80: **Full list+count refetch on success discards data the mutation already returned, and doesn't coordinate with the QR poll timer.**

`RECONNECT_WA_MANAGED_PHONE` already returns the updated row (`id, phone, status`), but it's discarded in favor of toggling `refreshList` here and at `PhoneManagement.tsx:150`, which forces `List.tsx`'s existing double round-trip (`refetchValues` + `refetchCount`) just to reflect one row. Separately, `ReconnectDialog.tsx`'s manual `refetch()` after logout (line 67) doesn't reset the 20s poll clock, so a poll due to fire shortly after can produce a redundant near-duplicate request. Consider updating the Apollo cache directly from the mutation response for the single-row case, reserving the full refetch for genuinely list-wide changes like bulk sync.

shijithkjayan (commented):
## Automated review summary

Reviewed the diff for `PhoneManagement.tsx`/`ReconnectDialog.tsx` and the surrounding wiring (menu/route/role checks, GraphQL queries/mutations). The PR is mostly additive and follows the repo's Apollo/i18n conventions correctly (await + try/catch for mutations, `useEffect` for query-derived state, matching i18n keys in both `en.json`/`hi.json`). Left inline comments on 8 points worth addressing before merge, ranked roughly by severity: a role-gating gap for custom "Dynamic" org roles, a cross-component status-classification inconsistency, a false-positive "reconnected" notification path, a couple of error-handling edge cases, a missing column style, and some refetch/duplication cleanup.

src/containers/WaGroups/PhoneManagement/PhoneManagement.tsx
```diff
@@ -0,0 +1,158 @@
...
+  phone: string;
+  label?: string | null;
+  status?: string | null;
+  lastStatusCheckedAt?: string | null;
+}
+
+export const PhoneManagement = () => {
+  const { t } = useTranslation();
+  const roles = getUserRole();
+  const isAdmin = roles.includes('Admin') || roles.includes('Glific_admin');
```
shijithkjayan: **This `Admin`/`Glific_admin` check is duplicated (slightly differently) in at least 2 other places** — `adminLevel` in `src/config/menu.ts` and a similar inline check in `WaGroups/GroupDetails.tsx/PhonesPanel/PhonesPanel.tsx`. None of them share a helper. This is the root cause behind the "Dynamic" role gap noted above: any future role change (new admin tier, renamed role) needs updating every copy, and missing one fails silently rather than erroring. Consider a shared `isAdminRole()`/`getUserRolePermissions()`-based check in `context/role.ts` instead.

src/containers/WaGroups/PhoneManagement/PhoneManagement.tsx
shijithkjayan: @akanshaaa19 We don't do dynamic roles, do we? If not, just mark this as resolved

src/containers/WaGroups/PhoneManagement/PhoneManagement.tsx
akanshaaa19: done

src/containers/WaGroups/PhoneManagement/PhoneManagement.tsx
akanshaaa19: done

src/containers/WaGroups/PhoneManagement/PhoneManagement.tsx
akanshaaa19: done, removed loading

src/containers/WaGroups/PhoneManagement/PhoneManagement.tsx
akanshaaa19: we dont

src/containers/WaGroups/GroupChatInterface/GroupChatInterface.tsx
```diff
@@ -43,6 +46,10 @@ export const GroupChatInterface = ({ collections }: GroupChatInterfaceProps) =>
     }
   }, [location]);
 
+  useEffect(() => {
+    cachedPhoneFilter = phonenumber;
+  }, [phonenumber]);
```
shijithkjayan, line 51: **Module-level `cachedPhoneFilter` leaks the phone filter across logout/login in the same browser tab.**

This is a plain module-scoped variable (not session/localStorage-backed), initialized once and mutated via the effect below. I checked `src/containers/Auth/Logout/Logout.tsx`: `handleLogout()` explicitly calls `resetRolePermissions()`, `clearListSession()`, `clearOrgEvalAccessCache()`, and `client.clearStore()` — a deliberate list of per-session UI state to reset — then does an **in-SPA** `navigate('/login', { replace: true })`, not a full page reload. Since JS module state survives a React Router navigation, a second user logging into the same browser tab after the first logs out will still see the first user's phone-filter selection applied in Group Chat. This contradicts the app's own established pattern of explicitly clearing this class of UI state on logout. Consider moving this into component state/`sessionStorage`, or adding a reset for it in `Logout.tsx`.

shijithkjayan (approved):
Overall LGTM. Address the 2 remaining comments and please add a screenshot/record of the pages

src/containers/WaGroups/GroupChatInterface/GroupChatInterface.tsx
akanshaaa19: handleLogout finishes with window.location.replace('/login') a full page navigation, not an in-SPA navigate so it reloads the whole page

akanshaaa19: and that seems like an overkill adding it to logout

src/containers/WaGroups/PhoneManagement/PhoneManagement.tsx
akanshaaa19: Reconnect is a manual low-frequency admin action on a small list (an org has a handful of managed phones), so the extra refetchValues + refetchCount round trip is negligible and keeps the server as the single source of truth rather than maintaining cache writes that can drift from the backend.

PR #4049, opened 2026-07-08
feat: Implement language variant management and UI enhancements

priyanshu6238:
## Summary

Closes: https://github.com/glific/glific/issues/5262

PR #4055, opened 2026-07-14
feat: Implement language version management 

### LINKED ISSUE #5262: [HSM Revamp]: Add Language & Fix and Resubmit Flows
**Part of Epic: #5258**
**Depends on: #5261**

## Summary

Two new dedicated pages — one for adding a translated language version of an existing template, and one for correcting and resubmitting a Meta-rejected template. Both are new routes that leave all existing files untouched.

## What problem does this solve?

**Add Language:** Today adding a Hindi translation means opening a blank form with zero reference to the English content. Staff copy-paste manually and often miss variables or buttons.

**Fix & Resubmit:** When Meta rejects a template, staff see a vague "REJECTED" status with no clear path to fix it. Most raise a support ticket instead of correcting it themselves.

**Example — Add Language:**
Staff opens `otp_verification` → clicks "+ Add Language". A new page shows the English body on the right as a collapsible reference strip: *"Your OTP is {{1}}. Valid for 10 minutes."* She types the Hindi version on the left. One click on "AI Translate" pre-fills it instantly. The button structure is already there — she only edits button labels. Submits. Done.

**Example — Fix & Resubmit:**
Staff sees `shipping_update → EN → Rejected`. She clicks "Edit & Re-apply". A page opens showing Meta's rejection reason: *"Message contains URL that is not whitelisted."* She fixes the URL, clicks "Reapply for Approval". No support ticket needed.

## Requirements

**Add Language** (`/template/:id/language/add`):
- New `AddLanguage.tsx` container — reads parent template id from URL params
- Two-card layout: Card 1 (template ID + tabbed language status), Card 2 (content form)
- Language tabs: Approved / In Progress / Rejected — empty tabs hidden; Rejected tab auto-selects when rejections exist
- Collapsible reference strip showing English source body text
- AI Translate button pre-fills body with machine translation in one click
- Shortcode pre-filled from parent and locked (not editable)
- Button structure inherited from parent and locked — only button labels are editable
- Tag locked (shared across all language versions)
- Submit CTA = "Submit for Approval"

**Fix & Resubmit** (`/template/:id/resubmit`):
- New `FixResubmit.tsx` container — reads template id from URL params
- Rejection reason strip shown at the top with Meta's rejection text
- Body + category fields editable; button structure + variable count locked
- Delete button available for rejected languages; disabled for pending
- Submit CTA = "Reapply for Approval"

- Both new routes registered in `AuthenticatedRoute.tsx`
- Existing `HSM.tsx` and all current routes completely untouched

## Acceptance Criteria

- [ ] Add Language page pre-fills shortcode from parent template and locks it
- [ ] Reference strip shows English body text and can be collapsed/expanded
- [ ] AI Translate pre-fills the body field in one click
- [ ] Language tabs hide when empty; Rejected tab auto-selects when rejections exist
- [ ] Fix & Resubmit page shows Meta rejection reason text
- [ ] Body and category are editable in resubmit mode; buttons and variable count are locked
- [ ] Submit CTA label is correct per mode ("Submit for Approval" vs "Reapply for Approval")
- [ ] Delete available for rejected; disabled for pending
- [ ] Existing `HSM.tsx`, `/template/add`, `/template/:id/edit` routes untouched

## Screenshots

Prototype: https://noopur-tech4dev.github.io/design-glific/HSM/add-hsm-language.html (same design repo)

priyanshu6238:
## Summary

Adds a Language versions feature to the HSM template create/edit page, letting NGOs manage all language variants of a template from one place:

- View — a read-only summary (/template-v2/:id/view) showing every language version of a template, grouped into Approved / In Progress / Rejected tabs.
- Add new language — pre-fills a draft from the anchor template's structure (category, buttons, media, etc.) so only the wording needs translating. The language picker excludes any language that already has a version, regardless of its status (Approved, Pending, or Rejected — not just Approved).
- Delete — removes a language version with confirmation; correctly falls back to the anchor if the deleted row was the one being previewed, and navigates back to the list if the anchor itself is deleted.
- Edit & Re-apply (Rejected/Failed variants) — opens an editable draft pre-filled with that variant's own content. On Save, the rejected record is deleted and the replacement created under the same shortcode/language — required because WhatsApp enforces uniqueness per (namespace, elementName, language), so the old rejected row has to be gone before the new one can be submitted.
- After creating or re-applying, the page stays put and refst's own query (the only one that returns real approvalstatus) so the new version shows in its correct tab instead of guessing.

Supporting changes:
- Extended the shared FormLayout component with two small,  props: beforeSubmit (async hook run before create/update)and headerHelp (overridable subtitle text).
- Fixed the Language dropdown for "Add new language" not retched (it only computed once on initial page load).
- Fixed HSM.helper.tsx's Call-to-Action button mapping not recognizing Gupshup's uppercase PHONE_NUMBER/URL button types, so previously-sbuttons weren't rendering their type picker correctly on ed
- Reorganized test mocks — moved inline, single-use Apollo mocks out of HSMV2.test.tsx into reusable factories in mocks/Template.tsx and mocks/Attachment.tsx.


src/containers/HSM/HSMV2/HSMV2.tsx
akanshaaa19, line 1: HSMs have no edit once submitted you can only view. So /:id/edit for an HSM is read-only, same as view. Collapse isEditing and isViewingVariant into a single isReadOnly/isView. That removes most of the fieldsLocked juggling on its own. and also look at how other components are implemented fieldsLocked was never needed

src/containers/HSM/HSMV2/HSMV2.tsx
```diff
@@ -106,9 +128,66 @@ export const HSMV2 = () => {
...
+  // local state (not a plain derived const) so deleting a variant from the
+  // Language versions card can drop it from the list immediately, without a
+  // refetch/renavigation.
+  const [familyVariants, setFamilyVariants] = useState<any[]>(location.state?.variants || []);
+  const isLanguageMode = Boolean(languageAnchorId) && !isEditing && !isCopyState;
+  const isAddRoute = isLanguageMode && !isViewRoute;
+  const isAddingLanguage = isAddRoute && addPagePanel === 'add';
+  const isViewingVariant = (isLanguageMode && isViewRoute) || (isAddRoute && addPagePanel === 'view');
+  const isEditingVariant = isAddRoute && addPagePanel === 'edit';
+  const fieldsLocked = isEditing || isViewingVariant;
```
akanshaaa19: 9 interdependent booleans (isCopyState/isViewRoute/isEditing/isLanguageMode/isAddRoute/isAddingLanguage/isViewingVariant/isEditingVariant/fieldsLocked) + a 3-value addPagePanel + a nested entityId ternary. Please replace with one mode enum: create | view | copy | addLanguage | reapply, computed once, everything derived from it.

akanshaaa19 (changes requested):
Copy already does fetch source → prefill → create new. Add-language and reapply are the exact same shape so they should reuse the copy path, not be separate modes with their own fetches. Just two navigate states into one "prefill-from-source" flow:

> - { mode: 'addLanguage', sourceId, excludeLanguageIds }
> - { mode: 'reapply', sourceId }

The only per-mode differences are small conditionals (shortcode locked, language picker filtered/locked, which submit mutation). Doing this lets you reuse FormLayout's existing getItemQuery prefill and delete languageAnchorData + previewData + the language-resolution effect the two extra fetches and the setStates race workaround I flagged last time all go away.

src/containers/Form/FormLayout.tsx
akanshaaa19, line 1: The client-side delete-then-create (delete the rejected variant, then create the replacement) isn't atomic — if the create fails, the rejected row is gone with no rollback. We shouldn't be reinventing a transaction on the client. Add a backend REAPPLY_TEMPLATE mutation that deletes + recreates in one transaction, and select it as the submit mutation for mode === 'reapply' (exactly how copy already swaps to CREATE_TEMPLATE). Then the new FormLayout.beforeSubmit prop isn't needed please revert the shared-component change.

akanshaaa19: can take this up later also

src/containers/HSM/HSMV2/HSMV2.tsx
```diff
@@ -106,9 +128,66 @@ export const HSMV2 = () => {
...
+
+  const isViewRoute = location.pathname.endsWith('/view');
+  const isEditing = Boolean(params.id && !isCopyState && !isViewRoute);
+  const languageAnchorId = isViewRoute ? params.id : location.state?.languageAnchorId;
+
+  const excludeLanguageIds = useMemo(() => location.state?.excludeLanguageIds || [], [location.state]);
+  // local state (not a plain derived const) so deleting a variant from the
+  // Language versions card can drop it from the list immediately, without a
+  // refetch/renavigation.
+  const [familyVariants, setFamilyVariants] = useState<any[]>(location.state?.variants || []);
```
akanshaaa19: useState(location.state?.variants || []) with no family fetch means a reload or direct-nav to /view shows empty tabs. Same point as #4049 — fetch the family by shortcode (one query) instead of threading the array through navigate(state).

src/components/UI/Pager/Pager.tsx
```diff
@@ -14,8 +14,10 @@ import { ColumnNames } from 'containers/List/List';
 import styles from './Pager.module.css';
 
 // `CollapseContent` carries a caller pre-rendered sub-row block (see the HSM
-// template list); `translations` is the default label+body JSON path.
-const removeDisplayColumns = ['recordId', 'translations', 'collapseContent', 'id', 'isActive'];
+// template list); `translations` is the default label+body JSON path;
+// `variants` is opt-in raw data some callers carry through for row actions
+// (e.g. HSM's "View"/"Add new language") without rendering it as a column.
+const removeDisplayColumns = ['recordId', 'translations', 'collapseContent', 'variants', 'id', 'isActive'];
```
akanshaaa19: The 'variants' addition to removeDisplayColumns only exists to carry the family array on list rows so it can be passed through navigation state. Once the view/add page fetches the family by shortcode (above), you don't carry variants on the row at all — so this shared-component change isn't needed. Revert it.

akanshaaa19 (changes requested):

src/containers/Form/FormLayout.tsx
```diff
@@ -73,6 +77,12 @@ export interface FormLayoutProps {
   getQueryFetchPolicy?: any;
   saveOnPageChange?: boolean;
   entityId?: any;
+  // separate from entityId: fetches from getItemQuery and runs setStates the
+  // same way entityId does, but doesn't make performTask treat this as an
+  // update, and doesn't suppress the isAttachment fresh-media-upload step —
+  // for flows that prefill from a source record while still creating a new
+  // one (e.g. HSMV2 add-language/reapply).
+  prefillId?: any;
```
priyanshu6238, line 81: note:

This is needed because prefillId lets us the  form to fetch and pre-fill from a source record without that source being treated as the row we are editing.mainly to fetch the messge_media instead of the raw url form the list

src/containers/HSM/HSMV2/HSMV2.tsx
```diff
@@ -275,6 +354,89 @@ export const HSMV2 = () => {
...
+    // dedicated /:id/view route.
+    if (!isViewRoute) {
+      setMode('view');
+      setAddPagePreviewId(variantId);
+    } else {
+      navigate(`/template-v2/${variantId}/view`, { state: { variants: familyVariants } });
+    }
+  };
+
+  const handleVariantCreated = async (data: any) => {
```
priyanshu6238, line 375: Here i am  using the formatLayout. This is just  fallback condition i have added for hard reload to load the data

src/containers/Form/FormLayout.tsx
akanshaaa19: Let's remove the Edit & Re-apply (rejected/failed variants) out of this PR for now. That needs a backend atomic reapply endpoint, which is a separate change.

 For this PR, please:
 - remove the reapply mode and the "Edit & Re-apply" action,
 - remove deleteRejectedVariantBeforeSubmit and the beforeSubmit={...} wiring, and
 - revert the beforeSubmit prop from FormLayout (nothing else uses it).

src/containers/HSM/HSMV2/HSMV2.tsx
akanshaaa19, line 1: Clicking "Add new language" currently prefills body/footer/buttons/media too, because it reuses copy's full prefill (prefillId → getItemQuery → setStates). For a translation we want everything blank except the element name and category (?).

akanshaaa19 (changes requested):

src/containers/HSM/HSMV2/HSMV2.tsx
akanshaaa19, line 1: in this file can you rearrange all variables properly? is looks very messy. have all usestates together, calculated variables together, same for queries, mutations and lastly all the functions

akanshaaa19:
there are too many comments- remove them most things are self explanatory. if something is not add it in PR description. keep the code clean

src/containers/HSM/HSMV2/HSMV2.tsx
priyanshu6238: But that's what we want, right? Since the value is present in the body and in other places , we can perform the auto-translation. The new UI also supports this. In the future, we'll add an Auto Translate button, so we'll need this behavior.

priyanshu6238: Done

src/containers/Form/FormLayout.tsx
priyanshu6238: Done

priyanshu6238:
> there are too many comments- remove them most things are self explanatory. if something is not add it in PR description. keep the code clean

Done

src/routes/AuthenticatedRoute/AuthenticatedRoute.tsx
```diff
@@ -131,6 +131,7 @@ const adminRoutes = (
     <Route path="template-v2" element={<HSMListV2 />} />
     <Route path="template-v2/add" element={<HSMV2 />} />
     <Route path="template-v2/:id/edit" element={<HSMV2 />} />
+    <Route path="template-v2/:id/view" element={<HSMV2 />} />
```
akanshaaa19: edit and view are same only right?

src/containers/HSM/HSMV2/HSMV2.tsx
akanshaaa19: we need auto-translate not prefilling, we add an extra step for the user this way. auto-translation can be done without filling the inputs. remove it

src/routes/AuthenticatedRoute/AuthenticatedRoute.tsx
priyanshu6238: No, it was used for the copy and edit the hsm form . See the copy icon on list page 
Screenshot 2026-07-16 at 11 13 50 AM
Screenshot 2026-07-16 at 11 14 11 AM

Screenshot 2026-07-16 at 11 14 22 AM

akanshaaa19: The existing HSM page (HSM.tsx) already handles this exact case with a single template/:id/edit route it renders read-only via isView, because an HSM can't be edited once submitted. That's the established pattern in this codebase.

there's no "edit" for HSMs, so these two routes mean the same thing, and the duplication introduces more variables extra code which is not needed. All of that exists only because there are two routes instead of one.

Please collapse to the pattern we already use: a single route for an existing template (read-only, since edit isn't supported), and /add for the create flows (blank / copy / add-language, distinguished by navigate state). Copy is a create-from-source, so it belongs on /add with the source in state — not on a /:id route.

priyanshu6238: Okay, I'll do it. However, it doesn't seem correct to me that the View page shows edit in the URL. 
Screenshot 2026-07-16 at 11 44 13 AM

akanshaaa19: keep any one of them. this will become easier to handle in HSM page also

priyanshu6238: Done

src/containers/HSM/HSMV2/HSMV2.tsx
priyanshu6238: Done

akanshaaa19, line 1: For add-language you're still fetching the entire anchor template via prefillId → GET_TEMPLATE, running setStates with all of it, seeding shortcode + category, and then if (mode === 'addLanguage') return to throw the rest away. That's patching the symptom, not fixing it.
Two concrete problems:
1. You fetch a whole template to use two fields and discard the other 80%. The ask was to not prefill add-language at all — seed shortcode/category from data we already have — not fetch everything and return out of setStates.
2. It's a redundant, circular fetch. The component also runs useQuery(FILTER_TEMPLATES, { filter: { shortcode: newShortcode } }) for the family which returns the anchor (shortcode + category) already. But that query is skip: !needsFamilyFetch || !newShortcode, and newShortcode only gets set by the prefill fetch's setStates. So it's prefill-fetch → sets newShortcode → unblocks a second fetch of the same template. The anchor is fetched twice and daisy-chained.▎
Please refactor instead of patching:
 - drop addLanguage from prefillId (no GET_TEMPLATE for this mode),
 - pass the anchor's shortcode in nav state (the list already has it — it's rendering the row) and key the family fetch off that,
 - seed newShortcode/category directly from the family/anchor, and
- delete the if (mode === 'addLanguage') return in setStates.

akanshaaa19: also prefillId isn't needed copy should reuse the HSM.tsx pattern. HSM.tsx already does copy without any prefillId: entityId = params.id drives the prefill, and updateItemQuery is swapped to CREATE_TEMPLATE so the "update" path creates a new row. HSMV2 only introduced prefillId because copy moved to /add (no route :id) and chose "no entityId → natural create" instead of reusing that pattern.

priyanshu6238: if we remove it attachment type will  breaks at copy  — attachment-type templates: entityId truthy skips the getMediaId step (!entityId check in FormLayout.tsx:220), which is what normally converts attachmentURL into a messageMediaId before submit. buildTemplatePayload (HSM.helper.tsx:279-281) also only strips attachmentURL when type === 'TEXT' or isEditing — and for copy, isEditing is false in both HSM.tsx and HSMV2.tsx (it's isReadOnly/its equivalent, and copy isn't read-only). So for an attachment-type copy with entityId set, attachmentURL (a raw string) survives into the payload untouched.

I checked the actual backend schema (glific/lib/glific_web/schema/session_template_type.ex:130-147) — session_template_input has no attachment_url field at all, only message_media_id. So that stray field either gets rejected outright by GraphQL input coercion or silently dropped, either way losing the attachment on the copy.

akanshaaa19: are you saying that currently attachments in copy is not working?

priyanshu6238: yes it seem  not woking

src/containers/HSM/HSMV2/LanguageVersionsCard/LanguageVersionsCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+  onTabChange,
+  showAddLanguage,
+  showDelete,
+  onView,
+  onAddLanguage,
+  onDelete,
+}: LanguageVersionsCardProps) => (
+  <div className={styles.TemplateDetailsCard}>
+    {/* the form below already has its own "Element name" field (locked
+        to the same shortcode) — showing it here too was a duplicate. */}
```
akanshaaa19: remove

src/containers/HSM/HSMV2/HSMV2.tsx
```diff
@@ -53,8 +68,21 @@ const SectionTitle = ({ title, helpData }: { title: string; helpData?: HelpDataP
   </div>
 );
 
+type Mode = 'create' | 'view' | 'copy' | 'addLanguage';
+
 export const HSMV2 = () => {
+  // ---- Hooks (routing / i18n) ----
```
akanshaaa19: remove the comments

akanshaaa19 (approved):
make sure you have tested everything, and remove all unnecessary comments

PR #4059, opened 2026-07-18
fix: improve error handling in FlowTranslation component

### LINKED ISSUE #5302: Surface and log Google Translate API errors (e.g. 403) instead of silently saving empty translations
### Context
 the *Interactive Template & Flow Auto-Translation Failure* incident (impact window ~2026-06-19 -> 2026-06-23; 9 orgs, 5006 failed translation calls). Root cause was a `403 API_KEY_SERVICE_BLOCKED` after the translation key was restricted to Gemini-only during Google's mandated key cleanup. Action  (fix the key restriction) is **Done (2026-06-23)**. This ticket covers the **error-handling defect** that made the failure invisible and delayed detection: the `403` was swallowed and never sent to AppSignal.

Raw error from Google Cloud Translation API:
```json
{
  "error": {
    "code": 403,
    "status": "PERMISSION_DENIED",
    "message": "Requests to this API translate method google.cloud.translate.v2.TranslateService.TranslateText are blocked.",
    "details": [{ "reason": "API_KEY_SERVICE_BLOCKED" }]
  }
}
```

### Problem
A hard external auth failure (`403 PERMISSION_DENIED`) was presented to users as a **successful save with empty translations**. The actual HTTP status and Google error body were discarded, and the error was recorded only via a raw `Logger.error` + a `translate_logs` row with `status = false` -- it was **never surfaced to AppSignal**, so no alert fired and detection was delayed by days.

### Where it is swallowed (located in code)
**1. `lib/glific/third_party/google_translate/translate.ex` -- status/body thrown away**
- `extract_translations/1` has a catch-all `defp extract_translations(_unexpected), do: {:error, "Failed to extract translations"}`. Any non-200 (including the `403 API_KEY_SERVICE_BLOCKED`) hits this clause, so the HTTP status and Google's error JSON are never read.
- `handle_response/3` `else` returns `{:error, "Invalid response #{inspect(response)}"}` -- and `inspect(response)` on a `Tesla.Env` would **leak the `X-Goog-Api-Key` header** (carried in `__client__`).

**2. `lib/glific/flows/translate/google_translate.ex` -- error never reaches AppSignal**
- `do_translate/3` on `{:error, error}` writes a `TranslateLog` with `status: false`, then calls raw `Logger.error(...)` and returns `""`. Because it uses `Logger.error` (not `Glific.log_error/2`), the failure **never goes to AppSignal**.

**3. `lib/glific/flows/translate/google_translate.ex` -- `translate/4` always reports success**
- The pipeline ends in `then(&{:ok, &1})`, so it **always** returns `{:ok, list}` even when every string is `""`. Callers (interactive templates `translate_quick_reply` / `translate_list` / `translate_location_request`, and flow Export/Import via `resolvers/flows.ex`) then persist the empty strings as a successful save.

### Proposed changes
1. **`translate.ex`**
   - In `extract_translations/1` / `handle_response/3`, pattern-match non-200 `Tesla.Env` and build an error that includes the **HTTP status and Google error body** (e.g. `"Google Translate API 403: API_KEY_SERVICE_BLOCKED -- Requests to this API ... are blocked"`), instead of the generic `"Failed to extract translations"`.
   - Replace `inspect(response)` with **`Glific.SafeLog.safe_inspect/1`** to avoid leaking the `X-Goog-Api-Key` header.

2. **`flows/translate/google_translate.ex` -- `do_translate/3`**
   - Report the `{:error, error}` path through **`Glific.log_error/2`** (Logger + AppSignal) instead of raw `Logger.error`, with org_id / source / target / status as AppSignal tags. Keep the `translate_logs` row. (Per repo convention: don't double-log -- `log_error` covers both Logger and AppSignal.)
   - Distinguish hard API errors (4xx/auth) from timeouts: keep the `""` retry fallback only for genuine timeouts, not for hard failures.

3. **`flows/translate/google_translate.ex` -- `translate/4`**
   - Stop unconditionally returning `{:ok, ...}`. When all (or a threshold of) strings failed with a hard error, return `{:error, reason}` so callers can avoid persisting blank translations and surface a real error to the user.

### Acceptance criteria
- [ ] A `403` / non-200 from Cloud Translation produces an error containing the **real HTTP status and Google reason** (`API_KEY_SERVICE_BLOCKED`) in logs and `translate_logs`.
- [ ] The failure is reported to **AppSignal** via `Glific.log_error/2` so an alert can fire -- not just a DB row.
- [ ] **No API key is leaked** in any log/error string (`safe_inspect` for Tesla envs).
- [ ] Hard translation failures are **not** silently saved as empty strings; the caller surfaces an error instead of a blank "success".
- [ ] Genuine timeouts still fall back to retry behavior (unchanged).
- [ ] Tests cover: 403 response -> structured error + AppSignal report + no blank save; 200 -> unchanged; timeout -> retry fallback.

priyanshu6238:
## Summary

Backend pr: https://github.com/glific/glific/pull/5412

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved auto-translation failure handling by showing the actual error details instead of a generic warning.
  * The translation dialog still closes after an auto-translation error, while the error message is reported to the user.
* **Tests**
  * Added coverage for auto-translation network error scenarios to verify the dialog behavior and displayed error messaging.


src/containers/Flow/FlowTranslation.test.tsx
```diff
@@ -65,6 +66,25 @@ describe('Testing Translation flows', () => {
...
+
+    await waitFor(() => {
+      expect(screen.getByText('Note')).toBeInTheDocument();
+    });
+
+    fireEvent.click(screen.getByText('Continue'));
+
+    await waitFor(() => {
+      expect(mockSetDialog).toHaveBeenCalledWith(false);
+      expect(errorMessageSpy).toHaveBeenCalled();
```
rvignesh89: Can we assert the exact error response struct so that we know what the backend sends?

rvignesh89 (approved):
Minor comment, otherwise LGTM

PR #4063, opened 2026-07-21
fix: centralise token renewal in TokenManager with axios/Apollo middleware

### LINKED ISSUE #4842: [Bug]: Unexpected Session Expiry and Automatic Logout while actively using Glific
### Summary

Experiencing unexpected logouts from the Glific instance. A popup appears stating “Session Expired”, and the user is logged out even while actively working on the platform.

following are some patterns/observations noticed.
- If a tab is idle for few hours, even if it works initially when we return, it logs out randomly after some time.
- If multiple tabs are open, and one tab gets session expired, the other tabs also get session expired.
- If a Glific tab is open for a long time and we come back to it, clicking on any module(flows, quick tools, chat....) shows “Session Expired.

### Steps to Reproduce

1. Login to Glific.
2. Open 2 - 3 tabs of the same Glific instance.
3. Keep the tabs idle for 1 - 2 hours.
4. Return and access any module

### Expected Behaviour

The system should not log the user out  while actively working.

### Actual Behaviour

1. A “Session Expired” popup appears.
2. The user is immediately logged out of Glific.
3. If multiple tabs are open, all of them get logged out as well.

### Acceptance Criteria

User should remain logged in while actively interacting with the platform.
Session should not expire unexpectedly during active usage.
Expiry in one tab should not automatically invalidate other active tabs.

### Screenshots

Image

rvignesh89:
## Why this PR

Fixes glific/glific#4842

Users were being logged out mid-session. The auth header + token renewal was driven by `setAuthHeaders()`, a global `fetch`/`XHR` monkey-patch that was only installed **while the flow editor was mounted**. On pages with no Apollo traffic to warm the token (notably `/analytics`), the token was never renewed — so the periodic embed-token refresh eventually ran with an expired token, got a 401, and kicked the user out.

This replaces the monkey-patch with proper middleware and one renewal authority.

## Changes

**What was broken → How it was fixed**

- **Renewal only happened while the flow editor was mounted** → Added `services/TokenManager` as the single renewal authority (`/session/renew`), plus `services/apiClient` (axios interceptors) and Apollo `authLink`/`errorLink`, so every authenticated call injects a token and renews on its own. Replaces `apollo-link-token-refresh` and the global monkey-patch.

- **Two tabs renewing at once logged one out** (the backend renewal token is single-use) → Renewal is serialized across tabs with the Web Locks API; the tab that waits adopts the winner's token instead of spending the rotated one.

- **A 2xx renew response missing `renewal_token` silently kept the spent token** (`setAuthSession` merges) → Validate the full token triplet before storing, and re-check validity after.

- **Any non-2xx `/renew` logged the user out**, so a proxy/WAF 403/404 logged out everyone → Terminal statuses narrowed to exactly `{400,401}`; everything else retries with backoff, no logout.

- **A hung `/renew` socket could freeze every tab (and logout) forever** → Timeouts on `apiClient` (30s) and the `/renew` POST (10s), plus a 45s lock-acquire abort that degrades to unlocked renewal.

- **On renewal failure the Apollo error link completed empty**, leaving `useQuery` loading forever and resolving awaited mutations to `undefined` → `errorLink` re-emits the original error.

- **`isAuthError` matched message text** (`"unauthorized"` / `"401"`), so ordinary permission denials triggered a renewal and burned the single-use token → Tightened to a real 401 / `UNAUTHENTICATED` code only.

- **Forced logout left the session in storage until the user clicked**, so `/renew` kept firing behind the dialog → `/logout/session` clears the session on mount.

- **The flow-editor library has no auth hook** → `services/flowEditorAuthShim`, a scoped `fetch`/`XHR` patch installed only while the editor is mounted and only for `FLOW_EDITOR_API` requests.

## Summary by CodeRabbit

* **New Features**
  * Added centralized authentication token renewal across REST, GraphQL, WebSocket, and flow editor requests.
  * Added proactive token validation, automatic retry after authentication failures, transient-error backoff, and cross-tab renewal coordination.
  * Added automatic redirection to session logout when renewal permanently fails.
  * Improved logout reliability by clearing local session data even when backend logout fails.

* **Bug Fixes**
  * Media validation now returns a consistent failure response instead of leaving requests unresolved.
  * Prevented repeated authentication retries and unnecessary WebSocket reconnections.

* **Documentation**
  * Updated authentication and token-renewal guidance.


src/services/apiClient.ts
```diff
@@ -0,0 +1,88 @@
...
+      }
+
+      // Guard against loops: this request gets exactly one renew-and-retry.
+      config!._retry = true;
+
+      // getValidAccessToken (not renewAccessToken): if another caller already replaced the token
+      // while this 401 was in flight, adopt it without spending the single-use renewal token again.
+      // Any ForcedLogoutError / TransientRenewalError simply rejects the caller (logout, if needed,
+      // was already emitted by TokenManager).
+      const token = await getValidAccessToken();
```
akanshaaa19, line 73: Use renewAccessToken() here as getValidAccessToken can use the exact token the server just rejected

src/config/apolloclient.ts
```diff
@@ -96,35 +66,50 @@ const gqlClient = (navigate: any) => {
...
-          navigate('/logout/session');
+    }
+
+    if (isAuthError(networkError, graphQLErrors) && !operation.getContext().__authRetried) {
+      operation.setContext({ __authRetried: true });
+
+      // getValidAccessToken (not renewAccessToken): if a concurrent caller already replaced the
+      // token, adopt it instead of spending the single-use renewal token again. `null` signals the
+      // renewal failed (forced-logout already emitted, or transient exhausted).
+      return fromPromise(getValidAccessToken().catch(() => null)).flatMap((token) => {
```
akanshaaa19, line 94: here also should use renewAccessToken() instead of getValidAccessToken() on this explicit post-401 retry path.

akanshaaa19 (approved):
overall looks good, left one minor comment. lets have this on staging for some time and test it

src/config/apolloclient.ts
rvignesh89: That's not what we want to do because renewal of token happens before the request is made so if an API returns a 401 then it likely only happens when in the backend the token is removed

PR #4072, opened 2026-07-24
fix: adjust label limit for Opt In and maintain option title limit

### LINKED ISSUE #3840: [Bug]: WhatsApp Forms UI, character lengths of options within selection components not right
### Summary

the UI prevents the user to enter option names more than 10 characters, but 10 characters is not the right option length. 
altering the form json and saving the changes enables the user to create the options. 

### Steps to Reproduce

- Create a form 
- add a selection component - ex- multicheck
- add options - and see the character limit of 10 characters, 
- try to add a string more than 10 characters and the string is cut short
- go to form json - and identify the option and edit the option in the json to add the larger string 
- apply changes to the json and toggle back to the UI to see that the option retains the larger string

### Expected Behaviour

the option fields should display and enforce the correct character limit as per meta documentation

### Actual Behaviour

string is cut short in UI, but modifying the form json lets the changes persist and the forms get published, because the actual character limit of options texts is longer. 

### Acceptance Criteria

find the right limit of characters for options in the forms and the input options should take in characters as per the correct limit.

### Screenshots

Image

akanshaaa19:
target issue is #3840 

## Summary by CodeRabbit

- **Improvements**
  - Updated character limits and validation for form labels, options, text fields, and button labels to match supported WhatsApp Flow requirements.
  - Added dynamic limits and character counters driven by the selected field type.
  - Refined selection constraints: “Opt In” label now caps at 120 characters, “Dropdown” at 20, and other cases at 30; option entries are capped at 30.
  - Updated text limits by type (e.g., Body, Caption, and date picker label behavior) and increased Button Label limit from 30 to 35.


priyanshu6238 (approved):
Looks good. Please fix the patch, as it's failing because many lines are only partially covered.

PR #4073, opened 2026-07-24
feat: switch org deletion to async deleteOrganization mutation

### LINKED ISSUE #5314: Switch frontend org deletion from deleteInactiveOrganization to deleteOrganization (async + notifications)
## Description

The frontend currently deletes organizations through the `deleteInactiveOrganization` GraphQL mutation. We should switch it to the `deleteOrganization` mutation, which deletes the org asynchronously via a background job and reports the result through in-app notifications, and **deprecate `deleteInactiveOrganization`**.

### Why switch

Both mutations are admin-only and ultimately erase all org data (`Erase.delete_all_organization_data/1`), but they behave very differently:

**`deleteInactiveOrganization`** → `Onboard.delete/2` (`lib/glific_web/resolvers/partners.ex:140`, `lib/glific/saas/onboard.ex:249`)
- Runs the **entire deletion synchronously inside the GraphQL request** — soft-deletes the org, then erases every related table inline.
- For organizations with large data volumes this can exceed the request/UI timeout, leaving the deletion half-done with no signal to the user.
- Provides **no notifications** — the caller only gets a synchronous `{:ok, org}` / `{:error, reason}`.
- Preconditions: requires `is_confirmed: true` and `organization.is_active == false`.

**`deleteOrganization`** → `Erase.delete_organization/1` (`lib/glific_web/resolvers/partners.ex:96`, `lib/glific/erase.ex:79`)
- Enqueues an **Oban background job** and returns immediately, so it never blocks or times out the UI regardless of org size. (This is the explicit purpose stated in the resolver/`Erase` docstrings.)
- The worker (`Erase.perform/1`, `lib/glific/erase.ex:118`) does soft-delete → `delete_all_organization_data` → emits an in-app **notification on both success and failure** (`send_success_notification` / `send_failure_notification`, category `Organization`, severity `:info` / `:critical`), plus structured `Logger` + AppSignal reporting.
- Precondition checked inside the worker: `organization.status == :ready_to_delete`.

In short, `deleteOrganization` is the newer, production-safe path: no UI timeouts and proper success/failure observability via notifications. The frontend should use it, and `deleteInactiveOrganization` should be deprecated.

## Things to reconcile during the switch

These are intentional differences between the two mutations that the FE change must account for:

1. **Different precondition.** The async path requires the org to be in `status: :ready_to_delete`, whereas `deleteInactiveOrganization` only required `is_active == false`. The FE flow must move the org to `:ready_to_delete` (e.g. via `updateOrganizationStatus`) before calling `deleteOrganization`, otherwise the worker rejects it with "Organization is not in deletable status".
2. **No `is_confirmed` argument.** `deleteOrganization` takes only `id` — there is no server-side confirmation guard. Keep the confirmation dialog on the client side before firing the mutation.

The mutation is fire-and-forget — it returns immediately and the background worker reports the outcome via `Organization` notifications. No loader or in-progress UI state is needed.

## Deprecating `deleteInactiveOrganization`

- Mark the field deprecated in the schema using Absinthe's `deprecate` macro (non-breaking — the field keeps working but is flagged in introspection/SDL), in `delete_inactive_organization` at `lib/glific_web/schema/organization_types.ex:399`, e.g.:
  ```elixir
  field :delete_inactive_organization, :organization_result do
    deprecate("Use deleteOrganization instead — it deletes asynchronously via a background job and reports success/failure through notifications.")
    arg(:delete_organization_id, non_null(:id))
    arg(:is_confirmed, non_null(:boolean))
    middleware(Authorize, :admin)
    resolve(&Resolvers.Partners.delete_inactive_organization/3)
  end
  ```
- Remove the field, its resolver (`Resolvers.Partners.delete_inactive_organization/3`), the `assets/gql/organizations/delete_onboarded.gql` asset, and its schema test (`test/glific_web/schema/organization_test.exs:300`) only **after** the frontend no longer references it. `Onboard.delete/2` can be reassessed at that point (still covered by `test/glific/partners/onboard_test.exs:212`).

## Acceptance Criteria

- [ ] Frontend organization-deletion flow calls `deleteOrganization` instead of `deleteInactiveOrganization`.
- [ ] The org is set to `ready_to_delete` status before the mutation is invoked (or the precondition mismatch is otherwise resolved).
- [ ] Client-side confirmation is retained before deletion is triggered.
- [ ] `deleteInactiveOrganization` is marked deprecated in the GraphQL schema via the `deprecate` macro.
- [ ] After the FE migration lands, remove `deleteInactiveOrganization` (field, resolver, `.gql` asset, schema test) as a follow-up.

shijithkjayan:
## What

Migrates the frontend organization-deletion flow from the synchronous `deleteInactiveOrganization` mutation to the async `deleteOrganization` mutation.

Closes glific/glific#5314

## Why

`deleteInactiveOrganization` runs the entire erase synchronously inside the GraphQL request, which can exceed the request/UI timeout for large orgs and provides no success/failure signal. `deleteOrganization` enqueues an Oban background job, returns immediately, and reports the outcome through in-app `Organization` notifications — the production-safe path.

## Changes

- **`OrganizationList.tsx`** — `deleteItemQuery` and the delete handler now use `DELETE_ORGANIZATION`. Before firing, the org is moved to `READY_TO_DELETE` status (via `updateOrganizationStatus`) to satisfy the async worker's precondition. Uses a dedicated `UPDATE_ORGANIZATION_STATUS` mutation instance so the status-dropdown's success toast doesn't leak into the delete path. Wrapped in `try/catch` with `setErrorMessage` per the repo's no-`onError` convention.
- Client-side name-confirmation dialog is **retained**.
- Toast copy updated to "deletion has been initiated" since the operation is now fire-and-forget (the background worker reports final success/failure via notifications).
- Removed the now-unused `DELETE_INACTIVE_ORGANIZATIONS` gql definition; repointed the `List.test.helper.ts` `orgProps` helper to `DELETE_ORGANIZATION`.
- Updated mocks and `OrganizationList` test for the new two-step flow.

## Acceptance criteria

- [x] FE deletion flow calls `deleteOrganization` instead of `deleteInactiveOrganization`
- [x] Org is set to `ready_to_delete` before the mutation is invoked
- [x] Client-side confirmation retained
- [ ] `deleteInactiveOrganization` deprecated in the GraphQL schema — **backend change** (`glific/glific`), tracked as a follow-up
- [ ] Remove `deleteInactiveOrganization` field/resolver/asset/test — **backend follow-up** after this lands

## Testing

- `OrganizationList` tests pass (2); `List` tests pass (9)
- `yarn format` clean, `tsc --noEmit` clean

## Summary by CodeRabbit

* **New Features**
  * Organization deletion now begins with a confirmation and status update before processing.
  * Users receive a notification when deletion has been initiated and will be notified when it completes.
  * Added Hindi translation for the deletion notification.

* **Bug Fixes**
  * Improved handling and display of deletion errors and warnings.

* **Chores**
  * Test artifact retention is now set to seven days.


.github/workflows/e2e-tests.yml
akanshaaa19, line 1: we can remove this not needed

akanshaaa19 (approved):

PR #4087, opened 2026-07-29
feat: implement design tokens and migrate hardcoded colors to CSS variables 

akanshaaa19:
### Summary

- Adds --app-color-* tokens to src/index.css + src/config/tokens.ts (primitive/semantic split, named by rank — green-primary, green-secondary, etc. — not an arbitrary numeric scale) and wires theme.tsx's hardcoded hex values to them. Only 8 tokens exist, each tied to a real call site today — grows incrementally as each feature migrates, not defined speculatively upfront.
- Separate --app-color-* namespace from the flow-editor widget's own --color-* vars in index.css — those are a vendored third-party theming API, not general app tokens, and reusing their names would risk restyling that widget by accident.
- Migrates ErrorPage.module.css as proof of concept — verified pixel-identical to the original hardcoded values.
- Adds scripts/migrate-hardcoded-colors.js — exact-match migration helper (colors, unlike spacing/font-size, don't need range-based snapping): parses the token map live from index.css so it can't drift, prefers a semantic name over the raw primitive when unambiguous, reports anything with no token yet instead of guessing.
- Ratchet baseline updated (hex-occurrence count reflects the one migrated file).

Test ### plan

- [x] Typecheck clean
- [x] Every substituted token value cross-checked against the original hardcoded hex (zero visual diff)
- [x] Migration script dry-run across components/UI: 47 safe replacements across 22 files with just the 8 existing tokens; correctly flags next-token candidates (#93a29b 16x, #fff/#ffffff 25x combined)
- [x] Unit tests pass (src/index.test.tsx)


akanshaaa19:
closing wrt https://github.com/glific/glific-frontend/pull/4109

PR #4088, opened 2026-07-29
fix: stop showing suspension modal to unauthenticated visitors

### LINKED ISSUE #4011: [Feature]:Display notifictation message for  Suspension/Force Suspension NGOs on Login Page
### Summary

Display a clear message on the Glific login page when an NGO account is in Suspension or Force Suspension (FS) state, explaining why the user cannot log in and what they should do next.

### Requirements

Currently, when an NGO account is in Suspension or Force Suspension, users attempting to log in receive a generic login failure without understanding the actual reason.

This creates confusion because:

Users assume they have entered incorrect credentials.
They repeatedly retry logging in.
They raise support requests before knowing that their organisation's account is suspended.
In many cases, the person attempting to log in may not be the registered billing/admin contact.

When a login attempt is made for an account whose organisation is in Suspension or Force Suspension, display an message instead of a generic login failure.

Example message:

Your organization's Glific account is currently suspended.

Please contact Glific Support team on support@glific.org for assistance.

### Acceptance Criteria

If the  NGO is in Suspension, or Force Suspension a message should be displayed instead of a generic login error.

### Screenshots

Image

Image

### Additional Notes

_No response_

shijithkjayan:
## Summary

Closes #4011.

The **Account Suspended** modal was fired on page mount in the shared `Auth` shell ([Auth.tsx](src/containers/Auth/Auth.tsx)), driven by the `status` field returned from the **public, unauthenticated** `POST /v1/session/name` endpoint. Because `Auth` is the shared shell for every auth screen, this modal popped up for **unauthenticated visitors** — before any login — on **Login, Registration, Confirm OTP, and both password-reset screens**.

As flagged by @rvignesh89 on the issue, we shouldn't reveal private account state (suspension) to unauthenticated users. The agreed approach is to surface suspension only **after** a login attempt with valid credentials.

## Changes

- **`Auth.tsx`** — remove the mount-time `checkOrgStatus(status)` effect and the `status` state. The `/v1/session/name` call now reads only the org **name** (for display), not the status.
- **`AuthService.tsx`** — remove the now-unused `checkOrgStatus` helper and its `setErrorMessage` import.
- **`Login.test.tsx`** — add a test asserting that loading the login page for a suspended org shows **no** modal.

## After-login behavior (unchanged)

Suspension is still communicated once credentials are validated, via the pre-existing `403` handling in `Login` `onSubmitLogin`:
- Valid credentials + suspended → backend `403` → suspension message shown.
- Invalid credentials → generic login error (no suspension leak).

## Testing

- `Login.test.tsx` (13) and `AuthService.test.ts` (10) pass.
- `tsc --noEmit` and `prettier --check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented suspension or private-account status messages from appearing to unauthenticated visitors during login.
  * Limited organization details displayed before authentication to the organization name.
  * Removed suspension notifications during the initial login page load.


src/containers/Auth/Auth.tsx
```diff
@@ -88,11 +86,14 @@ export const Auth = ({
       return;
     }
 
+    // Fetch only the organization name for display. The org's suspension status is
+    // intentionally not surfaced to unauthenticated visitors — doing so would leak
+    // private account state. Suspension is communicated after a login attempt, handled
+    // by the backend response in Login `onSubmitLogin`.
```
akanshaaa19, line 92: this comment can be removed, this is self explanatory

akanshaaa19 (approved):

PR #4090, opened 2026-07-30
feat: add spacing/typography design tokens with range-based migration

akanshaaa19:
### Summary

- Adds --app-space-* / --app-radius-* / --app-font-size-* tokens to src/index.css, defined in rem (not px) for browser zoom/accessibility — only steps with a real call site are declared (8 total), growing incrementally as each feature migrates.
- Wires the 4 in-scope hardcoded values in theme.tsx to these tokens.
- Migrates ErrorPage.module.css as proof of concept.
- Adds scripts/migrate-spacing-typography.js — unlike the color migration (exact-match only), spacing/font-size drift is accidental (12px/13px/14px all meaning "helper text"), so this snaps values to the nearest scale step, matching both px and rem in source files. A 2px-equivalent max-snap-distance cap prevents a sparse scale from silently forcing a distant value onto the wrong step.
- Prefers a semantic name over a raw step when real usage converges on one role (e.g. --app-font-size-helper), verified against actual CSS selector data, not assumption.

Test ### plan

- [x] Typecheck clean
- [x] Exact-match case verified (ErrorPage.module.css)
- [x] Snap + tie-break case verified (Dropdown.module.css, reverted after testing)
- [x] Max-snap-distance cap verified (rejects 8px rather than forcing it onto space-lg/12px)
- [x] rem-matching verified (Form.module.css, 9 declarations resolved)
- [x] Dry-run across components/UI: 140 declarations / 33 files resolve cleanly

## Summary by CodeRabbit

* **New Features**
  * Added app-wide design tokens for spacing, border radius, and typography.
  * Added a migration tool to convert eligible CSS values to design tokens, with preview and in-place update modes.

* **Style**
  * Updated error page spacing and typography to use design tokens.
  * Updated buttons, inputs, pagination, and responsive headings for consistent token-based styling.


akanshaaa19:
closing wrt https://github.com/glific/glific-frontend/pull/4109

PR #4091, opened 2026-07-30
feat: AI Evaluations V2.0

AkhileshNegi:
**Target issue is** https://github.com/glific/glific-frontend/issues/4098

## Summary
In this PR, adds the entry point for **AI Evaluations v2.0** behind a new org feature flag, so the
section can be enabled per-org while the page itself is still being built.

- **New feature flag `aiEvaluationV2Enabled`**.
- **Side drawer** — adds an "AI Evaluation v2" item under *AI toolkit*, pointing at `/ai-evaluation-v2`, restricted to manager-level roles. It's nested inside the existing `aiEvaluationsEnabled` block, so it shows only when **both** v1 and v2 flags are on.
- **Route** — `/ai-evaluation-v2` renders a lazy-loaded `AIEvaluation` container in the admin route tree.
- **Page** — placeholder screen with the heading and a "Coming soon" body; the real list lands in a follow-up.


src/containers/AIEvaluation/AIEvaluation.tsx
```diff
@@ -0,0 +1,11 @@
+import { t } from 'i18next';
+import { Heading } from 'components/UI/Heading/Heading';
+
+export default function EvaluationV2() {
+  return (
+    <>
+      <Heading formTitle={t('AI Evaluation v2.0')} />
+      <div data-testid="comingSoon">Coming soon</div>
```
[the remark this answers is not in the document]
Ayush8923: @coderabbit will change this in the next PR. This is just the initial draft, so I don't think we need to add this string to i18n at this stage.

src/config/menu.ts
```diff
@@ -203,6 +203,18 @@ const menus = (): Menu[] => [
               type: 'sideDrawer',
               roles: managerLevel,
             },
+            // ponytail: nested inside the aiEvaluationsEnabled block, so it needs v1 on too
+            ...(getOrganizationServices('aiEvaluationV2Enabled')
+              ? [
+                  {
+                    title: 'AI Evaluation v2',
```
akanshaaa19, line 209: can we keep the title just "AI Evaluation" instead of v2?

src/config/menu.ts
```diff
@@ -203,6 +203,18 @@ const menus = (): Menu[] => [
...
+                    path: '/ai-evaluation-v2',
```
akanshaaa19, line 210: and also the path? can we conditionally render the component based on the feature flag

akanshaaa19 (approved):
looks good, minor comments

src/config/menu.ts
Ayush8923: i have added the `AIEvaluationGuard` for this page and its related components.

src/config/menu.ts
Ayush8923: will rename this at the end and update the corresponding routes as well. we will incorporate this feedback in a future PR.

PR #4108, opened 2026-08-02
feat: Assistant Detail (Create/Edit) Page 

Ayush8923:
**Target issue is** https://github.com/glific/glific/issues/5483

## Summary
In this PR, added the assistant detail page for AI Evaluations v2. Clicking **edit** on a row, or **Create New Assistant**, now opens a dedicated page inside the evaluations section instead of jumping into the existing Assistants area.
- **New Page `AssistantDetail`** - header with back, assistant name, copyable assistant ID and "Publish & go live"; a card holding the version selector, evaluation health chip and the "live in your flows" note; then the tab strip (Persona & Prompt, Knowledge Base, Guardrails, Golden Q&A Evaluation, Try It Out `SANDBOX`). Each tab switches its panel and shows a "coming soon" placeholder for now.
- **Create vs edit** - `/ai-evaluation-v2/add` skips both queries entirely so nothing prefills, shows "New Assistant" and a `new assistant` pill, and hides the assistant ID, version selector, health chip and publish button.
- **Version selector** - driven by `GET_ASSISTANT_VERSIONS`. Lists every version newest first with its own status pill, defaulting to the live version and falling back to the latest. Status (`LIVE` / `not published`), the green treatment on the trigger.

Added so the AI Evaluation CSS modules contain zero hardcoded hex/px. Every value already
existed in the app; none had a token to reference.

**Colors — primitives**
- `--app-color-green-quinary: #a5d8bd` — border on the live-version pill.
- `--app-color-gray-quaternary: #e0e0e0` — hairline borders and dividers.
- `--app-color-gray-quinary: #191c1a` — near-black used for headings.
- `--app-color-white: #ffffff` — raised surfaces (cards, pills, menus).

**Radius** — only `lg` (12px) existed:
- `--app-radius-pill` (99rem, fully rounded)


akanshaaa19 (changes requested):
This is also a duplication of current Assistant create page, can you see if the old one can be used and you can conditionally render new things based on feature flag.

Ayush8923:
> This is also a duplication of current Assistant create page, can you see if the old one can be used and you can conditionally render new things based on feature flag.

@akanshaaa19 I think, we can’t directly reuse the current Assistant Create page because its create/edit UI differs significantly from the new Evaluation UI. The new Evaluation flow introduces a dedicated tab-based experience, so it requires a different layout and structure.
For refrence, here is the new Evaluation UI: https://drive.google.com/file/d/1gE50MVLNgW6BRkKWeJuFwlOQASPM6DRF/view?usp=drive_link

Also, from a code organization perspective, this belongs under the `AIEvaluation` domain. The existing Assistant module is an independent feature that we plan to deprecate. If we start conditionally rendering the new UI inside the old module behind feature flags, we will end up tightly coupling the new implementation with code that’s intended to be removed. That will make the eventual cleanup more difficult and increase the chances of regressions or breaking existing functionality.

By keeping the new implementation self-contained under AIEvaluation, we can simply remove the old Assistant module when it’s deprecated without affecting the new flow. This keeps the migration cleaner and avoids unnecessary dependencies between the old and new implementations.

akanshaaa19:
the header looks different use this components/UI/Heading 
image
and also make unsaved changes sentence case

akanshaaa19 (changes requested):
https://github.com/glific/glific-frontend/pull/4119#pullrequestreview-4879720443

Ayush8923:
> the header looks different use this components/UI/Heading image and also make unsaved changes sentence case

@akanshaaa19 as far as I can check, we can’t reuse the existing Heading component here because its styling is a bit different. If we use it, we will likely need to make changes to the component itself or override some of its styles, which I’d prefer to avoid for now.

I also checked the existing [Assistant page](https://github.com/glific/glific-frontend/blob/master/src/containers/Assistants/AssistantDetail/AssistantDetail.tsx#L142-L175), and although the UI is quite similar, I noticed that it has its own implementation instead of using the shared Heading component. So I followed the same approach here for consistency with the current implementation.

src/containers/AIEvaluation/AssistantDetail/AssistantDetail.tsx
```diff
@@ -0,0 +1,294 @@
...
+        >
+          <CopyIcon />
+          {assistant.assistantId}
+        </span>
+      )}
+
+      <div className={styles.VersionBar} data-testid="versionBar">
+        {isCreateMode || !selectedVersion ? (
+          <span className={styles.NewAssistantPill} data-testid="newAssistantPill">
+            {isCreateMode ? t('new assistant') : t('no version saved yet')}
```
akanshaaa19: can we make this sentence case

src/containers/AIEvaluation/AssistantDetail/AssistantDetail.tsx
```diff
@@ -0,0 +1,294 @@
...
+    const when = version.isLive ? t('published') : t('saved');
+    const timestamp = version.updatedAt ?? version.insertedAt;
+    return [version.description, timestamp ? `${when} ${dayjs(timestamp).fromNow()}` : null]
+      .filter(Boolean)
+      .join(' · ');
+  };
+
+  return (
+    <div className={styles.Page} data-testid="assistantDetailContainer">
+      <div className={styles.PageHeader} data-testid="heading">
```
akanshaaa19: can we use existing Header component here?

akanshaaa19 (approved):
two minor changes, otherwise looks good

src/containers/AIEvaluation/AssistantDetail/AssistantDetail.tsx
Ayush8923: Replied here @akanshaaa19 https://github.com/glific/glific-frontend/pull/4108#issuecomment-5213564332

src/containers/AIEvaluation/AssistantDetail/AssistantDetail.tsx
Ayush8923: This is already handle in this PR: https://github.com/glific/glific-frontend/pull/4118

src/containers/AIEvaluation/AssistantDetail/AssistantDetail.tsx
```diff
@@ -0,0 +1,294 @@
...
+        <div className={styles.HeaderLeft}>
+          <BackIcon
+            className={styles.BackIcon}
+            onClick={() => navigate('/ai-evaluation-v2')}
+            data-testid="back-button"
+          />
```
akanshaaa19: wrap this in <button aria-label={t('Back')}>

Ayush8923: Done. BTW I already handled this in this PR: https://github.com/glific/glific-frontend/pull/4119

PR #4114, opened 2026-08-03
fix: migrate design-system ESLint check to flat config

akanshaaa19:
## Summary
The Cypress-to-frontend move (already on master) bumped `eslint` to `^10.4.1` for its own flat-config-based linting. ESLint 10 doesn't support the legacy `--no-eslintrc` flag the design-system MUI-layering check relied on — breaking that check on **every** PR, not just design-system ones (see e.g. #4111).

Replaces `.eslintrc.design-system.json` with `eslint.design-system.config.mjs` (flat config, reusing the `typescript-eslint` package already added for Cypress), invoked via `--no-config-lookup --config` instead of `--no-eslintrc`. Same rule, same behavior.

## Test plan
- [x] `npx eslint --no-config-lookup --config eslint.design-system.config.mjs` against a known-bad file (`HSM.tsx`) — still flags it
- [x] Against a clean file — passes
- [x] Full-repo count matches the known 97-file baseline exactly
- [x] Typecheck clean, ratchet unaffected


eslint.design-system.config.mjs
```diff
@@ -0,0 +1,28 @@
...
+// never interferes with the app's own .eslintrc.json (airbnb) setup or cypress's own
+// flat config.
+const MUI_IMPORT_MESSAGE =
+  "Don't import @mui/material directly in src/containers/**. Use a shared component from " +
+  "src/components/UI (see src/components/UI/README.md) or, if it genuinely doesn't exist " +
+  'yet, add it there instead of inlining MUI here.';
+
+export default tseslint.config({
+  files: ['src/containers/**/*.ts', 'src/containers/**/*.tsx'],
+  ignores: ['src/containers/**/*.test.ts', 'src/containers/**/*.test.tsx'],
```
priyanshu6238, line 14: it's  still not make sense to me why we are ignoring in test case . do we allowed to use raw components here

priyanshu6238 (approved):

PR #4117, opened 2026-08-04
feat: implement design system checks to prevent raw MUI imports in containers

akanshaaa19:
## Summary by CodeRabbit

* **New Features**
  * Added automated design-system checks for code changes.
  * Added validation to detect newly introduced direct Material UI imports in container components.
  * Pull requests now verify design-system baselines and prevent new violations.

* **Chores**
  * Updated design-system guidance and validation scripts for clearer, more consistent checks.
  * Simplified linting configuration related to design-system enforcement.


scripts/check-new-mui-imports.js
```diff
@@ -0,0 +1,66 @@
...
+      newLineNumber = parseInt(hunkMatch[1], 10);
+      continue;
+    }
+
+    if (!inScope || !line.startsWith('+') || line.startsWith('+++')) continue;
+
+    if (MUI_IMPORT_RE.test(line)) {
+      violations.push(`${currentFile}:${newLineNumber} — ${line.slice(1).trim()}`);
+    }
+    newLineNumber += 1;
```
priyanshu6238, line 57: This numbering could become incorrect if someone imports a MUI component and then removes one or two lines above it. Can we handle that case here? It seems to me a bug

scripts/check-new-mui-imports.js
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Usage: node scripts/check-new-mui-imports.js <base-ref> [<head-ref>]
+ */
+import { execFileSync } from 'node:child_process';
+
+const MUI_IMPORT_RE = /^\+.*(?:from\s+|import\s*)['"]@mui\/material(?:\/|['"])/;
+const CONTAINERS_TSX_RE = /^src\/containers\/.*\.tsx?$/;
+const TEST_FILE_RE = /\.test\.tsx?$/;
+
+function main() {
```
priyanshu6238, line 18: The main function is hard to read here too many thing in main funtion. It would be better if we split the logic into smaller functions to improve readalibilty

scripts/check-new-mui-imports.js
```diff
@@ -0,0 +1,66 @@
...
+    if (MUI_IMPORT_RE.test(line)) {
+      violations.push(`${currentFile}:${newLineNumber} — ${line.slice(1).trim()}`);
+    }
+    newLineNumber += 1;
+  }
+
+  if (violations.length > 0) {
+    console.error('New raw @mui/material import(s) added in src/containers/**:\n');
+    for (const v of violations) console.error(`  - ${v}`);
+    console.error('\nUse a shared component from src/components/UI instead (see src/components/UI/README.md).');
```
priyanshu6238:
```suggestion
    console.error('\nUse components from src/components/UI instead of importing directly from @mui/material. See src/components/UI/README.md..');
```

scripts/check-new-mui-imports.js
```diff
@@ -0,0 +1,66 @@
...
+const TEST_FILE_RE = /\.test\.tsx?$/;
+
+function main() {
+  const [baseRef, headRef = 'HEAD'] = process.argv.slice(2);
+  if (!baseRef) {
+    console.error('Usage: node scripts/check-new-mui-imports.js <base-ref> [<head-ref>]');
+    process.exit(1);
+  }
+
+  const diff = execFileSync('git', ['diff', '--diff-filter=ACMR', '-U0', baseRef, headRef, '--', 'src/containers'], {
```
priyanshu6238: let use try catch block to safely handle it . the execFileSync function  can throw error  if Git fails:

priyanshu6238 (approved):
Approving it . please take a look at all the comments? A few of them are important and should be addressed, while the others are optimization suggestions that we can pick up later .

scripts/check-new-mui-imports.js
akanshaaa19: It's a 70-line script doing one thing- parse args, diff, scan, report. Splitting it into four functions will make it harder to read. not taking this one

scripts/check-new-mui-imports.js
akanshaaa19: newLineNumber resets from each hunk header (@@ -a,b +c,d @@ → start at c) and only advances on + lines, logging each one at the counter's value before incrementing. Since unified diff format guarantees that stripping all - lines from a hunk leaves the rest in exact new-file order, this stays correct no matter how additions and deletions interleave.

akanshaaa19: not a bug

PR #4123, opened 2026-08-06
fix: pre-select org's default language and other Template Library feedback fixes

### LINKED ISSUE #5515: [HSM Revamp Phase 2] Review Feedback for Template Library
1. In create page -> create from library -> the preview bot is overlapping 
2. In the template library modal -> can the default language selection be the default language of the org 
3. In the template modal -> Can the topic names not be opened, and be shown in closed state -> so the user can read through the topic names and go inside
4. The spacing and alignment of the button “template library” button seems a bit odd? Also can the order of the buttons be made as per the prototype 
5. Auto-translating -> then removing the auto-translation content -> autotranslating again is not filling the content 
6. Auto-translation -> not translating the variable content

priyanshu6238:
## Summary
- Pre-select the org's configured default language in the Template Library modal's language filter (falls back to "All languages" if the catalog has no entries in that language).
- Fix the language filter dropdown rendering behind the modal — its menu popped up at MUI's default z-index while this modal is forced above the WhatsApp Simulator panel (`z-index: 10001`); `Dropdown` now accepts an optional `menuProps` passthrough, and `DialogBox`'s `customStyles.root`/`customStyles.paper` now merge with existing classes instead of overwriting them.
- Collapse Template Library usecase groups by default so users can scan topic names before drilling in, instead of opening on a fully expanded list.
- Fix an `addLanguage` mode condition in `HSMV2` and clean up variable handling in `TemplateVariables`.

### DIFF
--- a/src/components/UI/DialogBox/DialogBox.tsx
+++ b/src/components/UI/DialogBox/DialogBox.tsx
@@ -124,7 +124,7 @@ export const DialogBox = ({
         container: styles.Dialogbox,
         paper: `${styles.DialogboxPaper} ${fullWidth && styles.FullWidth} ${customStyles?.paper || ''}`,
         scrollPaper: styles.ScrollPaper,
-        root: alwaysOntop ? styles.DialogboxRoot : '',
+        root: [alwaysOntop && styles.DialogboxRoot, customStyles?.root].filter(Boolean).join(' '),
       }}
       onClose={() => handleCancel()}
       aria-labelledby="alert-dialog-title"
--- a/src/components/UI/Form/Dropdown/Dropdown.tsx
+++ b/src/components/UI/Form/Dropdown/Dropdown.tsx
@@ -14,6 +14,7 @@ export interface DropdownProps {
   validate?: any;
   fieldChange?: any;
   fieldValue?: any;
+  menuProps?: any;
 }
 
 export const Dropdown = ({
@@ -25,6 +26,7 @@ export const Dropdown = ({
   form,
   fieldValue,
   fieldChange,
+  menuProps,
 }: DropdownProps) => {
   const { onChange, value, ...rest } = field;
 
@@ -53,6 +55,7 @@ export const Dropdown = ({
             classes: {
               paper: styles.Paper,
             },
+            ...menuProps,
           }}
           classes={{ outlined: styles.Outlined }}
           value={fieldValue !== undefined ? fieldValue : value}
--- a/src/containers/HSM/HSMV2/HSMV2.tsx
+++ b/src/containers/HSM/HSMV2/HSMV2.tsx
@@ -1,5 +1,5 @@
 import { useLazyQuery, useMutation, useQuery } from '@apollo/client';
-import { useEffect, useMemo, useRef, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { t } from 'i18next';
 import { useTranslation } from 'react-i18next';
 import { useLocation, useNavigate, useParams } from 'react-router';
@@ -163,7 +163,6 @@ export const HSMV2 = () => {
   const [tagId, setTagId] = useState<any>(location.state?.tag || null);
   const [variables, setVariables] = useState<any>([]);
   const [editorState, setEditorState] = useState<any>('');
-  const isAddLanguageModeRef = useRef(false);
   const [templateButtons, setTemplateButtons] = useState<
     Array<CallToActionTemplate | QuickReplyTemplate | WhatsappFormTemplate>
   >([]);
@@ -450,7 +449,6 @@ export const HSMV2 = () => {
   };
 
   const openAddLanguage = () => {
-    isAddLanguageModeRef.current = true;
     setMode('addLanguage');
     setAddPagePreviewId(null);
     setLanguageId(null);
@@ -760,11 +758,6 @@ export const HSMV2 = () => {
   useEffect(() => {
     if (location.state?.openAddLanguage && mode === 'view' && newShortcode && !anchorReference) {
       openAddLanguage();
-    }
-  }, [location.state?.openAddLanguage, mode, newShortcode, anchorReference]);
-
-  useEffect(() => {
-    if (isAddLanguageModeRef.current && !anchorReference) {
       return;
     }
     const bodyVariables = getVariables(body, variables);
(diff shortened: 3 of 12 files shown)


src/components/UI/Form/Dropdown/Dropdown.tsx
```diff
@@ -25,6 +26,7 @@ export const Dropdown = ({
   form,
   fieldValue,
   fieldChange,
+  menuProps,
```
priyanshu6238, line 29: 1- menuProps is an optional prop I added to the shared Dropdown component (src/components/UI/Form/Dropdown/Dropdown.tsx) purely to fix a z-index stacking bug in the Template Library modal — here's the full chain of reasoning:

The problem it solves

- The Template Library modal's Dialog needed to render above the WhatsApp Simulator panel, so its root got .AboveSimulator { z-index: 10001 } (Simulator sits at z-index: 10000).

- But the modal's language filter uses the shared Dropdown component, which wraps MUI's Select. MUI's Select doesn't render its options list inline — it portals a Menu/Popover to document.body, and that popover just uses MUI's theme default (theme.zIndex.modal, i.e. 1300) unless told otherwise. It has no awareness of what z-index the dialog it happens to live inside is using.

- So once the Dialog jumped to 10001, the dropdown's own options menu was still stuck at 1300 — meaning it rendered behind the dialog. Clicking the language filter opened a menu you literally couldn't see or click, because it was rendering underneath the modal.

Why an optional prop instead of hardcoding it

- Dropdown is a shared, generic component used all over the app — dozens of unrelated forms use it. I couldn't just bump its internal z-index directly; that would be a global change bleeding into every other dropdown in the app for a problem that only exists in this one modal.

src/components/UI/DialogBox/DialogBox.tsx
```diff
@@ -124,7 +124,7 @@ export const DialogBox = ({
         container: styles.Dialogbox,
         paper: `${styles.DialogboxPaper} ${fullWidth && styles.FullWidth} ${customStyles?.paper || ''}`,
         scrollPaper: styles.ScrollPaper,
-        root: alwaysOntop ? styles.DialogboxRoot : '',
+        root: [alwaysOntop && styles.DialogboxRoot, customStyles?.root].filter(Boolean).join(' '),
```
priyanshu6238, line 127: ## Why: 
- the Template Library modal needed its own z-index (.AboveSimulator, 10001) to render above the Simulator panel, but that's a one-off need specific to this modal — not something that should flip on via the existing alwaysOntop flag (which is a different, fixed z-index of 1501 meant for "above tooltips," used by other callers). So I extended the root slot to also accept a caller-supplied class through customStyles.root, letting TemplateLibraryModal pass customStyles={{ root: styles.AboveSimulator }} without touching alwaysOntop's existing behavior for every other DialogBox usage.

src/containers/HSM/TemplateLibraryModal/TemplateLibraryModal.test.tsx
akanshaaa19, line 1: Image

add some spacing between the button and the container

src/containers/HSM/HSMV2/HSMV2.tsx
```diff
@@ -504,23 +519,37 @@ export const HSMV2 = () => {
       }
       if (result?.body) {
         setBody(result.body);
-        setEditorState(result.body);
+        setEditorState('');
+        setTimeout(() => setEditorState(result.body), 0);
```
akanshaaa19, line 521: why are we clearing it first then updating the value

src/containers/HSM/HSMV2/HSMV2.tsx
```diff
@@ -163,6 +163,7 @@ export const HSMV2 = () => {
   const [tagId, setTagId] = useState<any>(location.state?.tag || null);
   const [variables, setVariables] = useState<any>([]);
   const [editorState, setEditorState] = useState<any>('');
+  const isAddLanguageModeRef = useRef(false);
```
akanshaaa19: We're using both a ref (isAddLanguageModeRef) and a state value (mode). is there nay reason why?

akanshaaa19 (changes requested):

akanshaaa19:
image

why is this gray? can we remove the bg color and add better border?

src/containers/HSM/HSMV2/HSMV2.tsx
```diff
@@ -504,23 +519,37 @@ export const HSMV2 = () => {
...
+        setTimeout(() => setEditorState(result.body), 0);
       }
       if (anchorReference.footer) {
         setFooter(result?.footer || '');
       }
-      if (result?.buttons?.length) {
+      const resultTexts: string[] = result?.buttons || [];
+      const translatedButtonTexts = resultTexts.slice(0, buttonTexts.length);
+      const translatedVariableTexts = resultTexts.slice(buttonTexts.length);
+      if (translatedButtonTexts.length) {
```
akanshaaa19, line 529: see if there is a better way to do this? relying on length is not the best way to do this

src/containers/HSM/HSMV2/HSMV2.tsx
priyanshu6238: Because sometimes, while translating again, the text that was left from the previous translation also gets appended to the top of the translated text.

src/containers/HSM/HSMV2/HSMV2.tsx
priyanshu6238: Removed: https://github.com/glific/glific-frontend/pull/4123/changes/f01cce9260fd36db5600503b56d72c8cd853c131 
Added by mistake

src/containers/HSM/TemplateLibraryModal/TemplateLibraryModal.test.tsx
priyanshu6238: Done: https://github.com/glific/glific-frontend/pull/4123/changes/f01cce9260fd36db5600503b56d72c8cd853c131

priyanshu6238:
> image
> why is this gray? can we remove the bg color and add better border?

Done: https://github.com/glific/glific-frontend/pull/4123/changes/f01cce9260fd36db5600503b56d72c8cd853c131

src/containers/HSM/HSMV2/HSMV2.tsx
priyanshu6238: Screenshot 2026-08-10 at 3 55 44 PM

The payload itself returns all the button translations in the same array under the button section, and we can distinguish  them using the length here.

akanshaaa19 (approved):

PR #4126, opened 2026-08-10
feat: Integrate List Models APIs

### LINKED ISSUE #5512: [Evals]: Integrate list-models endpoint
### Summary

Integrate Kaapi's `GET /api/v1/models` endpoint into Glific to fetch the live, active model catalog per provider, replacing whatever static/hardcoded model list currently backs the model selector (assistant config, prompt-generation, evaluations, etc.) so newly added providers/models on Kaapi's side show up without a Glific deploy.

### Requirements

- Call `GET /api/v1/models` (optionally filtered by `provider` as  `openai` only
- Can skip pagination for now since list will be less than 20 items while we deprecate old models
- Response `data` is a list of `ModelConfigPublic` objects — `provider`, `model_name`, `completion_type` (e.g. `text`, `stt`, `tts`) but will filter only `text` for assistant, `config` (per-model param schema, e.g. temperature bounds), `input_modalities`, `output_modalities`, `pricing`, `is_active`, `id`, `inserted_at`, `updated_at`; and `count` (total active models matching the filter).
- Surface at minimum `provider`, `model_name`, and `completion_type` to whatever picks the model list today; `config` can drive per-model param bounds (e.g. temperature min/max) instead of hardcoding them.
- Only active models are returned by Kaapi (sorted by `provider`, `model_name`) — no client-side filtering needed for `is_active`.
- Cache/refresh strategy should avoid calling this on every request; a periodic or on-demand refresh is acceptable given the model catalog changes infrequently.
- Existing hardcoded model list/behavior must keep working until this is wired in and verified, then can be removed.

### Acceptance Criteria

- [x] Glific calls `GET /api/v1/models` to populate the model selector with Kaapi's live, active model catalog.
- [ ] Response fields (`provider`, `model_name`, `completion_type`, `config`, `input_modalities`, `output_modalities`, `pricing`, `is_active`) are correctly parsed and surfaced to callers.
- [ ] Newly added Kaapi models/providers appear in Glific's model selector without a Glific code change or deploy.
- [ ] Error responses (`422`) are handled consistently with existing error-handling patterns.

### Additional Notes

Reference: https://api-staging.kaapi.ai/docs#/Model%20Config/Model%20Config-list_models — `GET /api/v1/models` (List Models)

Ayush8923:
**Target Issue is:** https://github.com/glific/glific/issues/5512

## Summary

- Replaces the hardcoded model catalogue in the AI Evaluation assistant editor with the backend's list. Adding or changing a model no longer needs a frontend release.

**What changed**
- Previously `assistantModels.ts` held a hardcoded `ASSISTANT_MODELS` array, each model's id, label, whether it was a "reasoning" model, which effort levels it accepted, and hardcoded `TEMPERATURE_MIN / TEMPERATURE_MAX` constants. That list is gone. `AssistantDetail` now queries `kaapiModels` and passes the result down to Persona & Prompt.
- Each model arrives with a JSON config describing the settings it accepts, so the tab renders itself from that rather than from an ahead-of-time classification:

Comes from the API | Was previously
-- | --
Which of temperature / effort / verbosity to show | A kind: 'standard' \| 'reasoning' flag
Effort levels offered | A fixed per-model efforts array
Temperature min / max | TEMPERATURE_MIN / TEMPERATURE_MAX constants
Helper text under each control | Hardcoded EFFORT_HINTS map
Default value per setting | A `defaultEffort` field

**Reasoning effort now persists**
- Separate from the API work, and a real bug: effort was never included in the `updateAssistant` payload, so whatever the user picked was discarded on save. The input now carries it, and `GET_ASSISTANT` selects effort so it comes back into the control on load.
- The payload is also gated on what the model declares, temperature is no longer sent to a model that doesn't accept one. Previously the control was hidden but the stale value was still submitted.

### Refrences:
- **Releated Backend PR:** https://github.com/glific/glific/pull/5513

## Summary by CodeRabbit

* **New Features**
  * Assistant configuration now loads available models dynamically.
  * Model-specific controls are shown based on supported settings.
  * Added support for reasoning effort and temperature configuration.
  * Models without adjustable settings now display an appropriate message.
* **Bug Fixes**
  * Assistant settings now preserve saved reasoning effort.
  * Save requests include only settings supported by the selected model.
  * Improved handling of incomplete or unavailable model data.
* **Localization**
  * Added messages for unsupported temperature controls and models without adjustable settings.


akanshaaa19 (changes requested):
verbosity is live in the UI and dirty-tracked, but never added to the input. Can we mirror the takesEffort pattern for takesVerbosity/verbosity, plus a test asserting it hits the mutation variables like the effort one does?

Ayush8923:
> verbosity is live in the UI and dirty-tracked, but never added to the input. Can we mirror the takesEffort pattern for takesVerbosity/verbosity, plus a test asserting it hits the mutation variables like the effort one does?

@akanshaaa19 i have completely removed the verbosity field from the UI. All models and their supported configuration fields are now dynamic, and Kaapi does not provide a verbosity field for any of the models.

The verbosity field that was present in the older UI was only part of the prototype and was not coming from Kaapi. I have also removed the related dead code.

akanshaaa19 (approved):

PR #4130, opened 2026-08-11
feat: Enable template library functionality across HSM components

### LINKED ISSUE #5536: Gate "Template Library" behind a feature flag (is_template_library_enabled)
## Summary
The "Template Library" button (Meta's pre-approved WhatsApp template catalog, browsable from both the HSM list page and the create-template page) currently ships unconditionally to every organization. Add a dedicated feature flag so it can be rolled out per-org instead of globally, matching how `is_prompt_generator_enabled` and `is_template_v2_enabled` are already gated.

## Motivation
Template Library is a newer surface (BSP partner API passthrough) that isn't ready for a blanket release. Gating it lets it be enabled selectively for pilot orgs and turned off instantly as a kill-switch without a deploy, consistent with the existing feature-flag rollout pattern used for other beta features.

### Backend (`glific/glific`) — FunWithFlags flag `is_template_library_enabled`, disabled by default
- `lib/glific/misc/flags.ex` — add to `init_fun_with_flags/1` so new orgs start with it off
- `lib/glific/partners/organization.ex` — new virtual field `is_template_library_enabled`
- `lib/glific/partners.ex` — wire into `fill_cache/1` and the `organizationServices` map (exposed as `template_library_enabled`)
- `lib/glific_web/schema/organization_types.ex` — expose on both `:organization` and `:organization_services_result` GraphQL types
- `lib/glific_web/schema/session_template_type.ex` — gate the `templateLibrary` query itself via `RequireFeatureFlag` middleware (defense-in-depth — not just a UI hide, since a disabled org shouldn't be able to hit the query directly either)
- `assets/gql/organizations/get_services.gql` — add the field
- Tests: `organization_test.exs` (services map assertion), `session_template_test.exs` (new test asserting the query is rejected with "not enabled" when the flag is off; existing `template_library` tests updated to explicitly enable the flag in setup)

### Frontend (`glific/glific-frontend`)
- `src/services/AuthService.tsx` — add `templateLibraryEnabled` to the `ServiceType` union
- `src/graphql/queries/Organization.ts` — add the field to `GET_ORGANIZATION_SERVICES`
- `HSMV2.tsx` (create-template page) and `HSMListV2.tsx` (template list page) — conditionally render the "Template Library" button and its modal on `getOrganizationServices('templateLibraryEnabled')`
- Tests: `HSMV2.test.tsx` / `HSMListV2.test.tsx` updated to enable the flag via `setOrganizationServices` in the tests that exercise the button/modal

## Rollout
Flag defaults to **disabled** for all orgs (existing and new). Enable per-org via `FunWithFlags.enable(:is_template_library_enabled, for_actor: %{organization_id: id})` (or equivalent admin/ops tooling) to pilot with specific organizations.

## Test plan
- [ ] `mix test test/glific_web/schema/session_template_test.exs test/glific_web/schema/organization_test.exs` passes
- [ ] Frontend: `HSMV2.test.tsx` and `HSMListV2.test.tsx` pass; `tsc --noEmit` and `prettier --check` clean
- [ ] Manual: with flag off, Template Library button/modal absent on both pages and `templateLibrary` GraphQL query rejected; with flag on, both work as before

priyanshu6238:
## Summary

- Add a feature flag for temp library

## Summary by CodeRabbit

* **New Features**
  * Added availability controls for the template library.
  * The template library button and modal now appear only when enabled.
  * The template library is available during HSM creation when enabled.
  * Improved placement and layout of template library controls.
  * Added conditional auto-translation controls when adding a language.

* **Bug Fixes**
  * Prevented unavailable template library features from appearing.


src/containers/HSM/HSMV2/HSMV2.tsx
```diff
@@ -182,6 +183,7 @@ export const HSMV2 = () => {
   } | null>(null);
   const [deleteTarget, setDeleteTarget] = useState<{ id: string; label: string } | null>(null);
   const [showLibrary, setShowLibrary] = useState(false);
+  const isTemplateLibraryEnabled = getOrganizationServices('templateLibraryEnabled');
```
akanshaaa19: can you move this to line 157? right now its between the states

priyanshu6238: Done: https://github.com/glific/glific-frontend/pull/4130/changes/b5d5f5b36c379fb19daf317adcd7b75bd98ee15f

src/containers/HSM/HSMV2/HSMV2.tsx
```diff
@@ -596,7 +598,10 @@ export const HSMV2 = () => {
       component: SectionTitle,
       name: '__sectionTemplateDetails',
       title: t('Template Details'),
-      action: mode === 'create' ? <TemplateLibraryButton onClick={() => setShowLibrary(true)} /> : undefined,
+      action:
+        mode === 'create' && isTemplateLibraryEnabled ? (
+          <TemplateLibraryButton onClick={() => setShowLibrary(true)} />
+        ) : undefined,
```
akanshaaa19:
```suggestion
      action: <TemplateLibraryButton onClick={() => setShowLibrary(true)} />,
      skip: mode !== 'create' || !isTemplateLibraryEnabled,
```

src/containers/HSM/HSMV2/HSMV2.tsx
```diff
@@ -596,7 +598,10 @@ export const HSMV2 = () => {
       component: SectionTitle,
       name: '__sectionTemplateDetails',
       title: t('Template Details'),
```
akanshaaa19, line 605: this is not the correct title for this?

src/containers/HSM/HSMV2/HSMV2.tsx
priyanshu6238: But if we use skip here , the whole section will disappear when sTemplateLibraryEnabled is not enabled.

src/containers/HSM/HSMV2/HSMV2.tsx
priyanshu6238: that title belong to the create hsm page first block and it is correct

Screenshot 2026-08-12 at 9 42 16 AM

akanshaaa19: then why is it in the template library component? same pattern is there for auto translate component. this is wrong.

also "Template Details" as a heading is redundant FormLayout already renders "Create a new HSM Template" right above it. Suggest dropping the "Template Details" SectionTitle field entirely and rendering TemplateLibraryButton directly in the component

akanshaaa19 (changes requested):
Also some other weird things i noticed:
- HSMV2.tsx:134: onChange is wired through inputProp: { onChange: (event) => setFooter(event.target.value) } instead of a top-level onChange. Input's documented contract  is onChange={(val: string) => ...} — a value, not an event — and every other field in this file (newShortcode, body, tagId) follows it. This only works today because MUI's InputBase happens to manually forward inputProps.onChange internally; inputProp is meant for native DOM attributes (maxLength, aria-*), not handlers. Please switch to onChange: (value: string) => setFooter(value) on the field descriptor and drop the inputProp wrapper — same behavior, matches the convention, doesn't rely on an MUI implementation detail.

- HSMV2.tsx:126 — on FooterField
This copies FormLayout.module.css's .FieldLabel verbatim (comment admits it). We already use CSS Modules composes for this elsewhere (SearchBar.module.css, DialogBox.module.css) can we do the same here instead of a hand-copied duplicate that can drift:
```
.FieldLabel {
  composes: FieldLabel from '../../Form/FormLayout.module.css';
}
```

- HSMV2.tsx:651-659: Can you please use skip here? instead of conditional rendering?

src/containers/HSM/HSMV2/HSMV2.tsx
priyanshu6238: I not added it in first place , but Noopur and Tejas shared feedback asking us to add all the headers for every section

priyanshu6238: why remove now?

priyanshu6238: Done: https://github.com/glific/glific-frontend/pull/4130/changes/92ec613bc32ac921116801d458376936e78c299d

priyanshu6238:
> Also some other weird things i noticed:
> 
> * HSMV2.tsx:134: onChange is wired through inputProp: { onChange: (event) => setFooter(event.target.value) } instead of a top-level onChange. Input's documented contract  is onChange={(val: string) => ...} — a value, not an event — and every other field in this file (newShortcode, body, tagId) follows it. This only works today because MUI's InputBase happens to manually forward inputProps.onChange internally; inputProp is meant for native DOM attributes (maxLength, aria-*), not handlers. Please switch to onChange: (value: string) => setFooter(value) on the field descriptor and drop the inputProp wrapper — same behavior, matches the convention, doesn't rely on an MUI implementation detail.
> * HSMV2.tsx:126 — on FooterField
>   This copies FormLayout.module.css's .FieldLabel verbatim (comment admits it). We already use CSS Modules composes for this elsewhere (SearchBar.module.css, DialogBox.module.css) can we do the same here instead of a hand-copied duplicate that can drift:
> 
> ```
> .FieldLabel {
>   composes: FieldLabel from '../../Form/FormLayout.module.css';
> }
> ```
> 
> * HSMV2.tsx:651-659: Can you please use skip here? instead of conditional rendering?

Done: https://github.com/glific/glific-frontend/pull/4130/changes/a6fa7f510279eb156bed4444d55bd8243c8e0250

src/containers/HSM/HSMV2/HSMV2.tsx
```diff
@@ -113,7 +114,12 @@ const AutoTranslateButton = ({
 // Same shortcut the HSM list page offers, surfaced here too so a user already
 // on the create page doesn't have to leave it to browse the catalog.
 const TemplateLibraryButton = ({ onClick }: { onClick: () => void }) => (
-  <Button variant="outlined" className={styles.AutoTranslateButton} onClick={onClick} data-testid="templateLibrary">
+  <Button
+    variant="outlined"
+    className={`${styles.AutoTranslateButton} ${styles.TemplateLibraryButtonRow}`}
```
akanshaaa19: can we give a generic name

src/containers/HSM/HSMV2/HSMV2.module.css
```diff
@@ -103,6 +104,13 @@
   height: 18px !important;
 }
 
+.TemplateLibraryButtonRow {
```
akanshaaa19: This is fragile since it's a hardcoded value from the top of the entire form (.CustomFormShell).

Since the button is already its own field (separate from SectionTitle), you can get the same right-aligned look without position: absolute at all wrap it in a row div and let flexbox do the alignment:


const TemplateLibraryButton = ({ onClick }: { onClick: () => void }) => (
  <div className={styles.TemplateLibraryButtonRow}>
    <Button variant="outlined" onClick={onClick} data-testid="templateLibrary">
      <GridViewIcon className={styles.AutoTranslateIcon} />
      {t('Template library')}
    </Button>
  </div>
);

.TemplateLibraryButtonRow {
  width: auto;
  display: flex;
  justify-content: flex-end;
}

Do the same for auto translate button and message content label. Also see and change if you have done this anywhere else. the label and component should match

akanshaaa19 (changes requested):

src/containers/HSM/HSMV2/HSMV2.tsx
```diff
@@ -99,7 +100,7 @@ const AutoTranslateButton = ({
 }) => (
   <Button
     variant="outlined"
-    className={styles.AutoTranslateButton}
+    className={`${styles.AutoTranslateButton} ${styles.SectionTitleAction}`}
```
akanshaaa19: either give this a generic name or remove `AutoTranslateButton`. its not adding much value here

akanshaaa19 (approved):

src/containers/HSM/HSMV2/HSMV2.tsx
akanshaaa19: @priyanshu6238

priyanshu6238: Done: https://github.com/glific/glific-frontend/pull/4130/changes/480d78d71d5c77c4e13e205e839df85b2f6889b0
Renamed to OutlinePillButton

PR #4138, opened 2026-08-14
fix: HSM status label, button casing, and translation source language

### LINKED ISSUE #5550: HSM translation fails with "Bad language pair: en|en" when target language is English
`translate_session_template/2` always assumes the anchor/source language of an HSM template family is English, regardless of what language the template was actually authored in. When a user tries to translate into English (or, symmetrically, translate an English-anchored template while some other resolution makes source == target), Google Translate is called with identical source and target locales and rejects the request outright.

## Log evidence
```
09:03:28.405 [warning] Translation failed for org 1 (English -> English): Google Translate API 400 : Bad language pair: en|en
```

## Root cause
`lib/glific/templates.ex` — `translate_session_template/2` called:
```elixir
GoogleTranslate.translate(texts, "English", language.label, org_id: organization_id)
```

Two problems in that one line:
1. Source language was hardcoded to `"English"` instead of using the template family's actual anchor language.
2. The destination was passed as `language.label` (a display name like `"Hindi"`), not `language.locale` (the actual code Google Translate expects, e.g. `"hi"`) — so any label happening to equal the source produces the same-language collision, and mismatched labels/locales could silently send wrong codes.

There was also no guard preventing a same-language translation request from reaching the Google Translate API at all — it should fail fast with a clear in-app error instead of surfacing a raw 400 from the provider.

## Steps to reproduce
1. Create/approve an HSM template in any non-English language (e.g. Hindi).
2. Use "Add new language" to translate that template into English.
3. Observe the request fails; server logs show `Translation failed for org <id> (English -> English)`.

## Expected behavior
- Translation should use the template's actual source language as `src`, not assume English.
- If source and target genuinely resolve to the same language, the API should return a clear, actionable error (e.g. `"Source and target language cannot be the same."`) instead of forwarding a doomed request to Google Translate.

## Fix (implemented on `fix/translation_issue_for_hsm`)
- Accept an optional `source_language_id` (GraphQL arg + context function), defaulting to English only when the caller doesn't specify one.
- Pass `locale` (not `label`) for both source and target to `GoogleTranslate.translate/3`.
- Add `validate_distinct_languages/2`, comparing `locale` (not `label`, since two language rows can share a label across different locales), to reject same-language translation requests before calling the provider.
- Added regression tests in `test/glific/templates_test.exs` covering: anchor-locale-as-source, same-locale rejection, and same-label/different-locale (non-)collision.

priyanshu6238:
## Summary
- Fixed the HSM template status help popover saying "Submitted" while the status filter dropdown said "Pending" for the same status — now both say "Pending", and the whole status-info block is properly wrapped in `t()` for translation (previously hardcoded English, not localized in Hindi).
- Standardized the "Sync HSM" button label casing (was `SYNC HSM` in one place, `Sync HSM` in another) and  removed a duplicate `"Sync HSM"` key that existed twice in both `en.json` and `hi.json`.
- Fixed a bug where the "Add new language" reference chips (body, footer, variables, buttons) always displayed **"English"** as the source language, even when the actual anchor template being translated from was in another language (e.g. Hindi/Marathi). This also caused Auto-translate to always request English → target instead of the real anchor language → target, which could produce mistranslated content or an outright "Auto-translation is not supported for English" error when translating into English from a non-English anchor.
- `HSMV2.tsx` now threads the anchor's actual `{ id, label }` through `anchorReference` and passes it to `SourceReferenceCard`, `FooterField`, `TemplateVariables`, and `TemplateOptionsV2` instead of a hardcoded string.
- The `translateSessionTemplate` mutation now sends `sourceLanguageId` (the anchor's real language) instead of relying on the backend defaulting to English. **Requires a companion backend change** (already applied separately) so the resolver accepts and uses `source_language_id` instead of hardcoding `"English"`.
- Refactored the repeated `mode === 'addLanguage' ? anchorReference?.x : undefined` checks (`anchorReference` is never cleared after the flow completes) into a single `activeAnchorReference` derived variable, removing four instances of duplicated logic.

## Summary by CodeRabbit

- **New Features**
  - Added source-language support for template references, variables, buttons, footers, and automatic translation.
  - Improved language-specific anchor handling and reference labels.
  - Added localized template status descriptions in English and Hindi.

- **Bug Fixes**
  - Updated template status wording from “Submitted” to “Pending.”
  - Standardized and translated HSM synchronization labels.
  - Improved fallbacks for untranslated reference content.

- **Tests**
  - Expanded coverage for language handling, translations, missing content, and draft preservation.


src/common/HelpData.tsx
```diff
@@ -119,19 +119,25 @@ export const templateStatusInfo: HelpDataProps = {
   heading: (
     <div>
       <p>
-        <b>Submitted:</b> The template is under review and can take up to 24 hours. This status indicates that the
-        template has not yet been approved or rejected.
+        <b>{t('Pending')}:</b>{' '}
```
akanshaaa19: @priyanshu6238 get the content reviewed by the PMs once

priyanshu6238: Done: https://discord.com/channels/717975833226248303/1516027052686311424/1538770921458901033

src/common/HelpData.tsx
```diff
@@ -119,19 +119,25 @@ export const templateStatusInfo: HelpDataProps = {
...
+        {t(
```
akanshaaa19: templateStatusInfo is a top-level const and JSX children evaluate eagerly, so these t() calls resolve once at import. It's also the only t() in a file of ~20 plain-string entries. Translate at render inside HelpIcon, or make heading a component.

src/i18n/hi/hi.json
```diff
@@ -268,6 +268,11 @@
   "Pending | Click Sync HSM to update the status": "लंबित | स्थिति अपडेट करने के लिए Sync HSM पर क्लिक करें",
   "Rejected": "Rejected",
```
akanshaaa19: "Approved", "Pending", "Rejected" still map to English. The descriptions are translated but the bolded labels aren't, so the popover renders half-Hindi.

src/common/HelpData.tsx
priyanshu6238: Done: https://github.com/glific/glific-frontend/pull/4138/changes/fbb6bd416585b0fad34893f7ad4ea7181795f372

src/i18n/hi/hi.json
priyanshu6238: Done: https://github.com/glific/glific-frontend/pull/4138/changes/fbb6bd416585b0fad34893f7ad4ea7181795f372

akanshaaa19 (approved):

PR #4147, opened 2026-08-18
feat: Evaluation V2 Feedback Updates

Ayush8923:
## Summary
- **Discard → Close:** the header action while a version has unsaved changes is now a secondary (outlined) button labelled Close instead of a plain text "Discard" link. Same behaviour: it asks before throwing the changes away.
- **Temperature is as a slider:** the plain number field is replaced by a slider paired with a number box, so a value can be dragged roughly or typed exactly. Both clamp to the model's declared range. Extracted as a shared [components/UI/Form/RangeSlider](https://github.com/glific/glific-frontend/pull/4147/changes#diff-6dd30b96a7feef563cf8f05a989a9031d5ce6c19ce244d6238cd8642958c811e) so raw MUI stays out of container code (the design-system ratchet enforces this).
- **"Technical details" → "Knowledge Base ID":** the section header now says what's actually inside it.
- **Download icon is easier to spot:** the muted grey icon is now a brand-coloured, slightly larger outlined download icon, so it reads as the primary action on a file row (delete stays muted).
- **SANDBOX tag removed** from the Try It Out tab.
- **Prompt can be expanded:** an expand icon sits in the bottom-right corner of the instructions box, matching the old assistant form. Clicking it opens a full-width **Edit system instructions** dialog with a large editor and **Cancel / Save**. No "Generate with AI" button inside the dialog. Unlike the old form, Cancel here genuinely discards the edit — the dialog works on a draft and only Save writes it back.
- Added the new paramater that name `duplicationFactor` in `aiEvaluations` query.

### Note:
- **Refrence Feedback Sheet:** https://docs.google.com/spreadsheets/d/1BnNlNpyzclqVqBka7Io4fYdTnlbAxO49siUC8zQI7YE/edit?usp=sharing


src/components/UI/Form/RangeSlider/RangeSlider.tsx
akanshaaa19, line 1: add tests for this

src/containers/AIEvaluation/AssistantDetail/Tabs/PersonaPrompt/PersonaPrompt.tsx
```diff
@@ -132,17 +149,15 @@ export const PersonaPrompt = ({ prompt, config, models, onPromptChange, onConfig
           {temperatureSpec && (
             <div>
               <div className={styles.FieldLabel}>{t('Temperature')}</div>
-              <Input
-                type="number"
-                placeholder=""
-                field={{ name: 'temperature', value: config.temperature, onBlur: () => {} }}
+              <RangeSlider
```
akanshaaa19, line 153: Number(config.temperature) || (temperatureSpec.min ?? 0) — || swallows the empty state, so the onClear on line 157 can never show an empty field. config.temperature === '' ? '' : ... (with RangeSlider accepting number | '') or a dedicated "unset" representation.

src/containers/AIEvaluation/AssistantDetail/Tabs/PersonaPrompt/PersonaPrompt.test.tsx
akanshaaa19, line 262: This test is named "clearing temperature reports an empty value rather than 0" and asserts onConfigChange receives '' — which it does. But renderTab never feeds the new config back in, so the rendered input is never re-evaluated. In the app AssistantDetail owns modelConfig and passes it down, so the box snaps to min. The test name asserts the behaviour that doesn't actually happen.

src/containers/AIEvaluation/AssistantDetail/Tabs/PersonaPrompt/PersonaPrompt.tsx
```diff
@@ -87,6 +88,22 @@ export const PersonaPrompt = ({ prompt, config, models, onPromptChange, onConfig
       <Input
         textArea
         rows={6}
+        endAdornment={
+          <span className={styles.Expand}>
+            <button
```
akanshaaa19: use icon button here

src/containers/AIEvaluation/AssistantDetail/Tabs/KnowledgeBase/KnowledgeBase.tsx
akanshaaa19: is this description corrrect since you changed title to knowledge base id

akanshaaa19 (changes requested):

Ayush8923:
@akanshaaa19 made the changes as per your comments, can you please re-review this PR.

akanshaaa19 (approved):
register RangeSlider in the ui readme

PR #4148, opened 2026-08-19
feat: post simulator messages to the authenticated endpoint

AmishaBisht:
## Summary

Frontend half of glific/glific#5578. Pairs with glific/glific#5555, which restricts the `/gupshup` webhook to Gupshup's published source IPs.

The simulator posted straight to the BSP webhook — the same URL Gupshup's own servers call:

```ts
export const GUPSHUP_CALLBACK_URL = `${GLIFIC_BACKEND_URL}/gupshup`;
axios.post(GUPSHUP_CALLBACK_URL, { type: 'message', payload: {...} })
```

That request is unauthenticated and lets the backend pick an organization from the request host. It also stops working the moment `/gupshup` is IP-restricted, because the request comes from the user's browser rather than from Gupshup.

### What changed

- `src/config/index.ts` — `GUPSHUP_CALLBACK_URL` becomes `SIMULATOR_MESSAGE_URL`, pointing at `/api/v1/simulator/message`.
- `src/components/simulator/Simulator.tsx` — both call sites (text at `sendMessage`, media at `sendMediaMessage`) post to the new endpoint with the session token attached, using the same `getAuthSession('access_token')` pattern as `Analytics.tsx`.

The auth header is built per request via a small `simulatorRequestConfig()` helper rather than captured once, since the token is renewed during a session.

Nothing else referenced `GUPSHUP_CALLBACK_URL`, so it is removed rather than left dangling.

### Why an endpoint change rather than keeping `/gupshup`

`/gupshup` stays exactly as it is and keeps serving real Gupshup traffic. Only the simulator moves off it. The alternatives were to let unauthenticated non-Gupshup callers through the IP filter, which reopens the vulnerability that #5555 exists to close, or to bolt authentication onto a webhook route and run two auth models on one endpoint.

As a side effect the simulator is now stricter than before: the organization comes from the caller's token instead of the `Host` header, so a user can only ever drive their own organization's simulator.

### Testing

- `vitest run src/components/simulator/Simulator.test.tsx` — 10 passed.
- `tsc --noEmit` clean.
- `prettier --check` clean on both changed files.

### Merge order

1. glific/glific#5578 — adds the endpoint (inert on its own).
2. **This PR** — starts using it.
3. glific/glific#5555 — starts rejecting unlisted IPs on `/gupshup`.

If #5555 lands before this one, the simulator is broken in the gap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated simulator message and media requests to use the simulator messaging endpoint.
  * Improved simulator request handling and error feedback when the connection is unavailable.
  * Added coverage for successful message and media submissions, as well as connection-loss notifications.


src/components/simulator/Simulator.tsx
```diff
@@ -107,6 +107,15 @@ const TimeComponent = ({ direction, insertedAt }: any) => (
...
 );
 
+// The simulator endpoint is authenticated, and the token is refreshed during a session,
+// so the header is built per request rather than captured once.
+const simulatorRequestConfig = () => ({
+  headers: {
+    Authorization: getAuthSession('access_token'),
+    'Content-Type': 'application/json',
+  },
+});
```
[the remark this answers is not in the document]
AmishaBisht: `services/apiClient` does not exist in this repository — there are no references to it in `src/`, and it is not on `master`. A shared authenticated client was added in #4063 and then reverted in #4131, so there is currently nothing to import.

The underlying concern was real though, and better than described. Testing this branch against a local backend showed every simulator message returning 401 and logging the user out. The request carried:

```
authorization: SFMyNTY.N2VhYTc0NDQ...R9U, SFMyNTY.N2VhYTc0NDQ...R9U
```

`AuthService.setAuthHeaders()` already patches `window.fetch` and `XMLHttpRequest.prototype.send` to attach the token, so the explicit header in the request config was a second copy; axios comma-joined them and the backend could not parse the result.

Fixed in 5c6fb2d26 by removing the explicit header, so the call is a bare `axios.post(url, body)`. That same patch also renews the token before sending when it has expired, which covers the renewal behaviour this comment asked for.

Verified in the browser: the flow editor preview posts to `/api/v1/simulator/message` with 200, the flow replies, and answering a prompt advances it.

src/components/simulator/Simulator.tsx
```diff
@@ -214,7 +214,7 @@ const Simulator = ({
       })
       .catch((error) => {
         // add log's
```
rvignesh89: remove these logs lines

rvignesh89 (approved):

src/components/simulator/Simulator.tsx
AmishaBisht: Removed in a2ada1cd3 — both `setLogs` calls are gone from the text and media catch handlers, along with the `// add log's` comment. The handlers now just set the disconnection state, which is what surfaces the failure to the user.

Worth noting the info line was also echoing the message body into the log on every failure (`sendMessageText:${sendMessageText} ...`), so this drops that too.

PR #4151, opened 2026-08-21
feat: Evaluation history export, live results and knowledge base downloads

Ayush8923:
## Summary

A batch of improvements to the AI Evaluation screens, focused on the History tab, the question-level results table, and knowledge base files.

### 1. History tab

- **Export CSV:** a button on the results table downloads every run as a spreadsheet: version, Golden Q&A set, duplication factor, status, and all four scores. The status gets its own column (rather than being mixed into the score) and dates are exported in full, so the file stays useful when opened later.
- **LIVE badge:** the version currently live in your flows is marked in the table, so you can tell at a glance which rows describe what your users are actually talking to. It reuses the same badge shown in the version bar, so it reads identically.

### 2. Question-level results: two ways to look at a run
- **Individual Rows** (the default, unchanged from before): one row per answer.
- **Group by Questions:** one row per question, with a column per attempt, so the answers to the same question sit side by side and can be compared directly.

### 3. The judge's reasoning is now readable

The judge writes an explanation for every mark it gives, and none of it was visible. Each score now carries an info icon; hovering it shows why that particular mark was given. This works in both views, against the right attempt in the grouped one.

### 4. Knowledge base file downloads

The download button on knowledge base files was disabled with a "not available yet" note. It now works, it fetches a fresh secure link and saves the file under its own name, showing a spinner on the button while it does. If the file cannot be fetched, the actual reason is shown rather than a vague message.

### 5. Runs are no longer blocked across versions
Starting an evaluation on one version used to lock the Run button for *every* version of that assistant, so we had to wait for it to finish before we could evaluate anything else, even though the two runs have nothing to do with each other. Runs are now tracked per version: you can evaluate version 2 while version 1 is still going. The one-at-a-time rule still applies within a single version, and the button's tooltip now says which version is busy.


src/containers/AIEvaluation/AssistantDetail/AssistantDetail.tsx
```diff
@@ -108,12 +109,36 @@ export const AssistantDetail = () => {
...
+      if (update.error) {
+        setErrorMessage(update.error, t('The prompt could not be improved'));
+        return;
+      }
+
+      const version = update.configVersion;
+      if (!version) return;
+
+      refetchVersions();
+      setSelectedVersionId(version.id);
```
akanshaaa19, line 142: does this switch the user's selected version on any org-wide improvePromptUpdated event. since this subscription takes no arguments, and configVersion and carries no assistant identity so when anyone in the org finishes an improve-prompt run, every open assistant-detail page jumps to a version that may belong to a different assistant

src/containers/AIEvaluation/AssistantDetail/Tabs/Evaluation/Evaluation.tsx
```diff
@@ -166,8 +190,12 @@ export const Evaluation = ({ assistantId, versionId, versionNumber, assistantNam
               color="primary"
               className={styles.RunButton}
               startIcon={<PlayArrowIcon />}
-              disabled={!versionId || runsInProgress}
-              title={runsInProgress ? t('An evaluation is already running. Wait for it to finish.') : undefined}
+              disabled={!versionId || versionRunInProgress}
+              title={
```
akanshaaa19, line 190: title on a disabled button doesn't render, so "which version is busy" never reaches anyone the body's claim about the tooltip isn't true. Fourth PR with this one, and this PR uses components/UI/Tooltip correctly at EvaluationScores.tsx:81. Same fix.

src/containers/AIEvaluation/AssistantDetail/Tabs/KnowledgeBase/KnowledgeBase.tsx
```diff
@@ -72,9 +73,37 @@ export const KnowledgeBase = ({
...
+      return;
+    }
+
+    const signedUrl = data?.getFile?.signedUrl;
+    if (!signedUrl) {
+      setNotification(t('This file has no download link yet. Try again in a moment.'), 'warning');
+      return;
+    }
+
+    downloadFile(signedUrl, data?.getFile?.filename || file.filename);
```
akanshaaa19: downloadFile sets both target="_blank" and download (common/utils.ts:112). Browsers ignore download cross-origin, so a signed GCS/S3 link opens a tab rather than saving, and the filename passed here is decorative unless the bucket sets Content-Disposition: attachment.

check this once

src/containers/AIEvaluation/AssistantDetail/Tabs/KnowledgeBase/KnowledgeBase.tsx
```diff
@@ -72,9 +73,37 @@ export const KnowledgeBase = ({
...
   const fileInputRef = useRef<HTMLInputElement>(null);
 
   const [uploadFile] = useMutation(UPLOAD_FILE_TO_KAAPI);
+  const [fetchFile] = useLazyQuery(GET_KNOWLEDGE_BASE_FILE, { fetchPolicy: 'network-only' });
+
+  const handleDownload = async (file: KnowledgeBaseFile) => {
+    setDownloading(file.fileId);
+
+    const { data, error } = await fetchFile({ variables: { fileId: file.fileId } });
+    setDownloading(null);
```
akanshaaa19: setDownloading(null) is only on this path. If fetchFile throws at the link level instead of returning { error }, this row's spinner never stops and there's no recovery without a reload. try { … } finally { setDownloading(null) }.

akanshaaa19 (changes requested):

src/containers/AIEvaluation/AssistantDetail/Tabs/KnowledgeBase/KnowledgeBase.tsx
Ayush8923: so i checked this, and this is a real limitation, not an oversight. `downloadFile` helper is used across the app, including the Golden QA export.

for cross-origin files, the browser may ignore the download attribute, so whether the file downloads or opens in a new tab depends on the bucket’s Content-Disposition setting.

akanshaaa19 (approved):

PR #4156, opened 2026-08-27
fix: Evaluation UI Feedback Updates

### LINKED ISSUE #4154: [Feature]: Evaluation results — metric order, inline grouped scores, consistent metric naming
## Summary

Three readability/consistency issues on the evaluation run result screen (AI Toolkit → Assistant → Evaluation tab):

1. **"Adherence to knowledge base" sits in the middle of the score breakdown.** It can sometimes be `Not scored in this run.`, so an unscored row currently interrupts the two rows that always have a score. Moving it last keeps the scored checks together and the "may be absent" one at the bottom, where it is easier to skim past.
2. **In "Group by Questions", each metric renders on its own line inside the answer cell** (`Ground Truth 3.0 ⓘ` / `Prompt 5.0 ⓘ` / `Knowledge Base — ⓘ` stacked vertically). Three stacked lines per answer make rows very tall; a single line would fit far more questions on screen.
3. **Metric names are inconsistent between the two halves of the screen** — both in wording and in casing:
   - Overall score breakdown: `Adherence to ground truth`, `Adherence to knowledge base`, `Adherence to prompt` (sentence case, `Adherence to` prefix kept).
   - Per-row / per-answer scores: `Ground Truth`, `Prompt`, `Knowledge Base` (Title Case, prefix stripped).

   These come from two different sources, which is why they drifted:
   - Panel labels are hardcoded in the frontend — `METRIC_ORDER` in [EvaluationResult.tsx:31](src/containers/AIEvaluation/AssistantDetail/Tabs/Evaluation/EvaluationResult/EvaluationResult.tsx:31).
   - Table labels are derived from the backend `score.name` (`"Adherence to Ground Truth"`) via `traceMetricNames` → `shortMetricName` ([evaluation.ts:149](src/containers/AIEvaluation/utils/evaluation/evaluation.ts:149)).

   So editing the `METRIC_ORDER` strings alone will drift again the next time the backend changes a name.

## Screenshots

Image

Ayush8923:
## Summary
These feedbacks are covered in this PR:

1. **Evaluation progress on the tab** — a live dot now sits on the Golden Q&A Evaluation tab while a run is going, so you can move to another tab and still know it's running. It follows the selected version, so a run on one version doesn't mark the others.
2. **Clearer wait message** — the in-progress card now says the run takes about 10 minutes and that you're free to leave the page, instead of only saying there's nothing to show.
3. **Supported file formats are visible upfront** — the knowledge base now lists what it takes (PDF, DOC, DOCX, TXT, MD, HTML, CSV) and the 20MB per-file limit right under the Add files button, so an upload doesn't fail for reasons you couldn't have known.
4. **Files with capital extensions upload** — a file named `REPORT.PDF` was being turned away while `report.pdf` went through; the extension case is now handled.
5. **Knowledge base empty state** — reworded to "No files added yet. Add files for the assistant to use when answering questions."
6. **Each evaluation check explains itself** — an info icon beside every check (ground truth, prompt, knowledge base) says what the judge is measuring there.
7. **Checks reordered** — knowledge base now reads last, since it's the one most runs leave unscored.
8. **Judge marks read as one line** — in Group by Questions the per-answer marks sit in a single strip under the answer with a marker before each, instead of a tall stack, and the info icons are properly centred.
9. **"Go Live" button** — renamed from "Publish & Go Live", and when it's greyed out, hovering now tells you why (already live, still building, or failed).
10. **"Model & Prompt" tab** — renamed from "Persona & Prompt".
11. **Golden Q&A wording** — the word "set" is gone everywhere; it now reads "Golden Q&A", "Manage Golden Q&A", "Add Golden Q&A", and so on.
12. **Evaluation Health column** — the scoring guide moved behind an info icon (it used to pop up on the whole header) and now states the correct 0–5 range instead of 0–1.
13. **Health pills line up** — every band sits at the same width, so the column reads as one stack instead of pills of three different sizes.
14. **Last Updated column removed** from the assistant list; the list still shows newest first.

### Note
- **Refrence Feedback Link:** https://docs.google.com/spreadsheets/d/1BnNlNpyzclqVqBka7Io4fYdTnlbAxO49siUC8zQI7YE/edit?gid=0#gid=0


src/containers/AIEvaluation/AssistantDetail/Tabs/Evaluation/EvaluationResult/EvaluationResult.tsx
```diff
@@ -98,7 +112,7 @@ export const EvaluationResult = ({
           <div className={styles.PendingTitle}>{t('Evaluation in progress')}</div>
           <div className={styles.PendingNote}>
             {t(
-              "This runs in the background, so there's no progress to show — the result appears here once it completes. You can switch tabs or leave this page."
+              'This runs in background, estimated time of about 10 mins. You can switch the tabs or leave this page. The result will be displayed here once the evaluation process is completed'
```
akanshaaa19:
```suggestion
              'This runs in the background, estimated time of about 10 mins. You can switch tabs or leave this page. The result will be displayed here once the evaluation process is completed'
```

src/containers/AIEvaluation/AssistantDetail/Tabs/KnowledgeBase/KnowledgeBase.tsx
```diff
@@ -161,7 +171,7 @@ export const KnowledgeBase = ({
   };
 
   const handleAddFiles = async (event: React.ChangeEvent<HTMLInputElement>) => {
-    const selected = Array.from(event.target.files ?? []);
+    const selected = Array.from(event.target.files ?? []).map(withLowercaseExtension);
```
akanshaaa19, line 174: shouldn't this be handled in the backend? frontend should be able to pass any file name

src/containers/AIEvaluation/AssistantList/AssistantList.tsx
```diff
@@ -213,35 +205,28 @@ export const AssistantList = () => {
...
-            'Scored 0–1 by our automated judge. 0–0.3 = Needs Improvement. 0.3–0.6 = Needs Refinement. 0.6–1 = Good.'
-          )}
-          placement="top"
-        >
-          <span className={styles.HealthHeader}>{t('Evaluation health')}</span>
-        </Tooltip>
+        <span className={styles.HealthHeader}>
+          {t('Evaluation health')}
+          <Tooltip
+            title={t('Scored 0–5 by our automated judge. 0–1 = Needs improvement. 2–3 = Could improve. 4–5 = Good.')}
```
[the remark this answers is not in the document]
akanshaaa19: @Ayush8923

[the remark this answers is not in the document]
Ayush8923: i think the current tooltip title is fine because we have defined the scoring consistently the same way across the evaluation flow. If we define the scoring differently here, it could make the UI inconsistent and potentially create confusion for users.

src/containers/AIEvaluation/AssistantDetail/Tabs/KnowledgeBase/KnowledgeBase.tsx
Ayush8923: i am not completely sure whether this is already being handled on the backend, but I think it’s still good to validate it on the frontend as well. this gives us an additional layer of validation and security, rather than relying entirely on the backend.

akanshaaa19 (approved):
looks good, last thing It says "0–1 = Needs improvement. 2–3 = Could improve. 4–5 = Good", but scoreBand is >= 4 good, >= 2 okay, else bad and scores are decimals, so a 1.5 or 3.5 falls in no band the tooltip names. A tooltip whose job is explaining the scale shouldn't get the scale wrong. CodeRabbit flagged the same thing

Ayush8923:
> looks good, last thing It says "0–1 = Needs improvement. 2–3 = Could improve. 4–5 = Good", but scoreBand is >= 4 good, >= 2 okay, else bad and scores are decimals, so a 1.5 or 3.5 falls in no band the tooltip names. A tooltip whose job is explaining the scale shouldn't get the scale wrong. CodeRabbit flagged the same thing

thanks for suggestion but we are following the same consistency at the product level as well, so I don’t think we should change it only here. The scoring is not provided at the decimal level for individual judge metrics. decimal can be shown for the top-level overall score, but the individual row-level judge metrics are scored using the defined whole-number ranges.

PR #4157, opened 2026-08-27
feat: Semantic major/minor Version for assistant config

Ayush8923:
## Summary
In this PR, Assistant versions are now shown as **semantic version labels** (`1.0`, `1.1`, `2.0`) instead of a single running number (`Version 1`, `Version 2`, `Version 3`).

matches how versions are actually created: saving an assistant produces a **minor** bump, publishing promotes it to the **next major** and makes it live. Until now the UI collapsed both into one counter, so a reader could not tell a saved draft apart from a published version by looking at the number.

- **Assistant list —** the "Live version" column reads `Version 2.0` instead of `Version 2`.
- **Assistant detail —** the version pill, the version dropdown, the "is live in your flows" note, and the In `Progress / Failed` badges all use the `major.minor` label.
- **Try it out —** "Testing Version 2.1 · sandbox only, real users stay on Version 2.0".
- **Evaluation tab —** the run panel, the run dialog, the result summary, and the History table all label runs with the full version.
- Exported evaluation CSV carries the `major.minor` label in its Version column.
- **AI Evaluations list —** the assistant sub-line reads `Assistant / Version 2.0`, and its link points at that exact version.
- **Create-evaluation dropdown —** options read `Assistant (Version 2.0)`.
- Evaluation run names generated for the backend now embed the full label (`assistant_v2_0_set_name_…`).

### Note
- **Releated PR:** https://github.com/glific/glific/pull/5591


src/containers/AIEvals/AIEvaluationList/AIEvaluationList.tsx
```diff
@@ -108,27 +108,30 @@ interface NameCellProps {
...
+    minorVersion?: number | null;
     assistant?: { id?: string | null; name?: string | null } | null;
   } | null;
 }
 
 const getName = ({ name, goldenQa, assistantConfigVersion }: NameCellProps) => {
   const assistantName = assistantConfigVersion?.assistant?.name;
-  const versionNumber = assistantConfigVersion?.versionNumber;
+  const versionLabel =
+    assistantConfigVersion?.majorVersion != null
```
akanshaaa19, line 120: This is the fourth place that turns a version into a label: configVersionLabel() at evaluation.ts:201, inline at AIEvaluationCreate.tsx:157, here, and the backend's own versionLabel string. They don't agree — you guard a null minor with ?? 0, the other two render 2.undefined. If the guard is needed here it's needed there too, or nowhere if versionLabel from the server is authoritative. Right now the same version can print two different ways in two views. Pick one owner: either always use the server's versionLabel, or route everything through configVersionLabel.

src/containers/Assistants/VersionPanel/VersionPanel.tsx
```diff
@@ -41,9 +43,12 @@ interface VersionPanelProps {
   onSelectVersion: (version: AssistantVersion) => void;
   onRefetchSelect?: (version: AssistantVersion) => void;
   refetchTrigger?: number;
-  initialVersionNumber?: number;
+  initialVersionLabel?: string;
 }
 
+const compareVersionsDesc = (a: AssistantVersion, b: AssistantVersion) =>
```
akanshaaa19, line 49: This is compareVersionsDesc from utils/assistantVersions.ts, re-declared verbatim in the same PR that adds that util for exactly this. Same on line 25, where AssistantVersion is redeclared with the same three new fields as types/assistantType.ts:36. Import both

src/containers/AIEvaluation/utils/assistantVersions.ts
```diff
@@ -0,0 +1,7 @@
+import type { AssistantVersion } from 'containers/AIEvaluation/types/assistantType';
+
+export const compareVersionsDesc = (a: AssistantVersion, b: AssistantVersion) =>
+  b.majorVersion - a.majorVersion || b.minorVersion - a.minorVersion;
+
+export const isNewerThan = (version: AssistantVersion, previous: AssistantVersion | null) =>
```
akanshaaa19, line 6: you have to invert a descending comparator to convince yourself < 0 means newer. There's one call site (AssistantDetail.tsx:188); a direct major-then-minor comparison would say what it means.

akanshaaa19 (approved):
minor comments, otherwise okay

src/containers/AIEvals/AIEvaluationList/AIEvaluationList.tsx
Ayush8923: this split is based on what each API currently returns. `versionLabel` is only available on `assistantVersions`, while the evaluation APIs only provide `majorVersion` and `minorVersion`. so for now, the evaluation views have to build the version label themselves.

Ayush8923: The `?? 0` is also because of the different type definitions. here, `minorVersion` can be null, while the evaluation types define it as a number.

I can move the version-label formatting into a shared `configVersionLabel()` function so we don’t have the same logic in multiple places.

src/containers/Assistants/VersionPanel/VersionPanel.tsx
Ayush8923: these are actually two different modules. `containers/Assistants/` is the existing `Assistants` UI, while `containers/AIEvaluation/` is the new AI Toolkit UI that will eventually replace it.

I intentionally kept them independent so that we can remove the old module later without having to deal with dependencies from the new module.

Ayush8923: the types are also slightly different, so they can’t be directly shared without some changes or type casts. If we want to share this logic, I think the better approach would be to move it to a neutral shared utility rather than keeping it inside `AIEvaluation/utils`.

src/containers/AIEvaluation/utils/assistantVersions.ts
Ayush8923: this helper keeps the version sorting and version comparison consistent. Both use the same `compareVersionsDesc` logic, and it’s already covered by unit tests.

Inlining the major/minor comparison here would mean duplicating the same logic again.

PR #4158, opened 2026-08-31
feat: AI assistant flow updates

Ayush8923:
## Summary
- AI Assistant menu now opens the new assistant flow, moved it to `/assistants` route
- Removed the separate "AI Evaluation v2" menu item
- Disabled the "Create AI Evaluation" button on the old AI Evals page (runs start from the assistant's Golden Q&A tab now)
- Added confirmation modal on Go Live, with a nudge to run an evaluation if the version was never evaluated
- Added category wise model listing in the model dropdown (Recommended / All models / To be deprecated) with badges
- New assistant now defaults to the recommended model
- Model dropdown width now matches the field width
- Added "Last used" tag on the Golden Q&A used in the last run, and it is selected by default
- Added info icons on Ground truth, Knowledge base, Prompt and Overall columns in the History table
- Highlighted the Overall column and made the score columns equal width and centered
- Made the assistant list column spacing even
- Grouped the split i18n strings into full sentences
- Remove the `Not Published` tag from the Version dropdown. 

### Note:
- **Related Backend PR:** https://github.com/glific/glific/pull/5651


src/containers/AIEvaluation/AssistantDetail/Tabs/Evaluation/Evaluation.tsx
```diff
@@ -84,13 +92,24 @@ export const Evaluation = ({
...
 
+  const lastScoredRun = versionRuns.find((run) => !isRunInProgress(run) && !isRunFailed(run)) ?? null;
+
+  useEffect(() => {
+    onLastRunChange?.(lastScoredRun);
+  }, [lastScoredRun, onLastRunChange]);
+
+  useEffect(() => {
+    if (openRunSignal > 0 && sets.length > 0) setRunOpen(true);
+  }, [openRunSignal, sets.length]);
```
akanshaaa19, line 111: sets.length in the deps makes this fire repeatedly. runRequest only increments, so once the publish nudge is used openRunSignal > 0 is permanently true and any change to sets.length reopens the Run dialog — add a Golden Q&A later and it springs open on its own. Track the last handled signal in a ref so a consumed signal stays consumed.

src/containers/AIEvaluation/utils/assistantVersions.ts
```diff
@@ -5,3 +5,10 @@ export const compareVersionsDesc = (a: AssistantVersion, b: AssistantVersion) =>
 
 export const isNewerThan = (version: AssistantVersion, previous: AssistantVersion | null) =>
   !previous || compareVersionsDesc(version, previous) < 0;
+
+export const nextPublishLabel = (versions: AssistantVersion[], selected: AssistantVersion, hasLiveVersion: boolean) => {
```
akanshaaa19, line 9: This reimplements the backend's bump rule client-side and shows the prediction as a promise in the dialog title. It drifts if the server rule changes, and two admins publishing at once both compute the same highestMajor + 1. Better to have the backend return the prospective label, or drop the number from the title and show the real one after the mutation.

akanshaaa19 (changes requested):
lets not remove the feature flag, change that and conditionally render the component based on the feature flag

src/containers/AIEvaluation/AssistantDetail/Tabs/Evaluation/Evaluation.tsx
Ayush8923: this signal only fires after someone clicks `Run an evaluation` in the publish dialog, and the dialog is meant to open there. The reopen you describe needs a Golden Q&A to be added later in the same session. so far, we have not come across any this type of such edge case during testing, and there is no bug or breaking change.

src/containers/AIEvaluation/utils/assistantVersions.ts
Ayush8923: i don’t think we need to implement anything for this right now, and we have not come across any such cases during testing either. If it comes up, we’ll need to coordinate with the backend team, and we can implement it later if needed.

akanshaaa19 (approved):
please test this properly

PR #4170, opened 2026-09-04
feat: Golden Q&A total question sizes

Ayush8923:
## Summary
`goldenQas` now returns the full golden QA type, so `totalItems` is available. How big a dataset is shows in three places:
- the Golden Q&A dialog list — `Added 15 hours ago · 120 questions`
- the dataset view — `… asks these 120 questions.`, taken from the file that was read back, with the
  server's count standing in while it loads or if it cannot be read at all
- the run evaluation dropdown — so it is clear how big a run is about to be before starting it

Sets uploaded before the count existed report nothing rather than `0 questions`.

## Note:
- **Related PR:** https://github.com/glific/glific/pull/5698
- **Feedback Sheet:** https://docs.google.com/spreadsheets/d/1BnNlNpyzclqVqBka7Io4fYdTnlbAxO49siUC8zQI7YE/edit?gid=0#gid=0


akanshaaa19 (approved):
minor comment

Evaluation.test.tsx — the legacy case is covered with totalItems: null, but total_items is NOT NULL DEFAULT 0, so every pre-count Golden Q&A actually arrives as 0. The test passes either way: strip the > 0 from goldenQaItemCount and it still goes green, because typeof null === 'number' is false regardless. Worth adding a 0 row so the guard is actually pinned — null is still schema-legal (the field isn't non_null), so keep both.

PR #4171, opened 2026-09-07
feat: work the web channel from the inbox

### LINKED ISSUE #5659: [Epic] Web Channel — Phase 1
## Web Channel — Phase 1

Phase 0 (#5443) covered the prototype. **Phase 1 is the production build**: a browser-based
messaging channel an NGO can switch on for itself, reusing Glific's flow engine, contacts and
staff inbox unchanged.

Beneficiaries open an NGO-branded web page, authenticate with an OTP delivered over WhatsApp,
and hold a conversation that appears in the same staff inbox as WhatsApp — driven by the same
flow engine.

### Design references

- **Technical design (current)** — [`plans/web-channel/`](https://github.com/glific/glific/tree/web-channel-tech-design/plans/web-channel) on the `web-channel-tech-design` branch: `tech-design.md`, `api-auth-design.md`, `custom-ui-design.md`, `ngo-api-integration.md`, `frontend-hosting-decision.md`
- **Omnichannel messaging design** — https://claude.ai/code/artifact/381a0eea-5253-4748-987b-3780ecfe78fe
- **Web Channel API (partner integration proposal)** — https://claude.ai/code/artifact/f1ef87be-0644-49c6-bc0b-d6bcff13a2a4
- **Cross-channel user scenarios** — see the shared Google doc; stories are referenced here by `US*` number

> The two artifact links are access-controlled and will 404 for anyone they have not been shared with.

### What changed since Phase 0

Phase 0 locked several decisions that Phase 1 supersedes. Recording them so the difference is not rediscovered in review:

| Phase 0 | Phase 1 |
|---|---|
| No org theming | Per-org theming, set in admin settings and loaded at runtime |
| Real SMS OTP | **OTP delivered over WhatsApp** — SMS may not be enabled in time, and Glific already sends OTPs this way |
| One active flow per contact across all channels | Per-channel flow state via `flow_contexts.channel` |
| No capability-gating framework | `ChannelCapability` registry, so a channel that cannot render a feature degrades rather than errors |
| Manual enablement, no Settings UI | Per-org feature flag plus theme fields in Settings |

### Scope of Phase 1

**In:** channel discriminator and per-org flag; per-org theming; OTP auth over WhatsApp; sending
text, audio, image, file and location from the web; the staff inbox showing and answering web
conversations; flows replying on web; AppSignal metrics; BigQuery columns; per-org rollout.

**Out, and deliberately:** `contact_identities` and phone-less contacts; custom UI nodes; partner-minted
JWTs; load testing. Each is recorded with its reasoning in the technical design.

### Sub-tickets

Listed in dependency order. Remaining tickets will be added as they are cut.

- [ ] #5660 — Channel discriminator on `messages` and `flow_contexts`, behind a per-org feature flag
- [ ] #5661 — Per-org theming, plus the Playwright e2e harness
- [ ] #5662 — Authenticate contacts by OTP delivered over WhatsApp
- [ ] #5663 — Web channel sends text, audio, image, file and location
- [ ] #5664 — Admin inbox shows and filters web channel messages
- [ ] #5706 — Every flow entry point takes a channel, defaulting to WhatsApp
- [ ] #5707 — A trigger cannot start a web-only flow
- [ ] #5711 — Web channel configuration: settings page, brand assets, media bucket and per-org signing key
- [ ] #5713 — Separate web channel opt-in from WhatsApp opt-in
- [ ] #5719 — Flows reply on the web channel
- [ ] #5720 — Organisation profile page in the widget, and letting a contact edit its own name
- [ ] #5721 — Product and system metrics for the web channel
- [ ] #5722 — Exploratory testing of the web channel, and fixing what it finds
- [ ] #5725 — Ad-hoc fixes: no login when the channel is disabled, and Playwright coverage for login
- [ ] #5768 — Per-channel conversation counters: All, Unread, Not replied, Not Responded

_More to follow: admin replies with offline delivery, delivery status, BigQuery columns, and per-org rollout._

**Phase 2** is #5702 — contact identities, custom UI nodes and Exotel inbound, each deferred from Phase 1 with a recorded reason.

## Dependency graph

```
#5660 ─┬─▶ #5661
       ├─▶ #5662 ─┬─▶ #5663 ─▶ #5664
       │          ├─▶ #5711
       │          └─▶ #5713
       └────────────┘
```

- **#5660** blocks everything — it is the schema and the flag.
- **#5661** (theming + harness) branches off and runs in parallel with the auth/messaging line.
- **#5662 → #5663 → #5664** is sequential: authenticate, then send, then surface in the inbox.
- **#5711** is everything an organisation configures before the channel can be switched on — the settings page, brand assets, the media bucket, and the per-org signing key that hardens what #5662 issues — per-org signing keys replacing the single installation-wide key — and can run in parallel with #5663. It must land before a second organisation is enabled, because until it does, cross-tenant isolation of the socket token rests on a claim check rather than on cryptography.
- **#5719** connects the two ends the messaging tickets left open: web inbound never reaches the flow
  engine, and a flow'"'"'s outbound messages never carry a channel. It depends on #5706 for the entry
  points and on the outbound adapter built alongside #5713, and it owns deriving `flows.channels` —
  without that derivation nothing stops a template-only flow being started on a web contact.
- **#5720** extends #5661'"'"'s branding endpoint with the About-the-organisation fields #5711 collects,
  and reinstates contact rename — removed from #5663 in review because it rode on the socket and
  intercepted every outbound message to keep a header in sync. It needs both of those tickets first.
- **#5721** runs alongside the feature tickets rather than after them — a rollout that cannot be read
  per organisation cannot be judged per organisation. It carries one open decision: the widget has no
  analytics at all today, and adding any to a page beneficiaries use is a privacy call, not a
  technical one.
- **#5722** is the last gate before the first real organisation. Several things the channel depends on
  cannot be reached by any test — bucket CORS, an OTP arriving on a real phone, a socket over a real
  network, two organisations at once — and every defect found so far that mattered was found this way.
- **#5713** corrects what #5662 records — a web login currently opts the contact in to WhatsApp — and
  adds `contact_histories.channel`. It runs in parallel with #5663 and must land before the first real
  beneficiary signs in, because every login until then writes a WhatsApp consent record that was never
  given.

### LINKED ISSUE #5664: Admin inbox shows and filters web channel messages
**Epic:** #5659
**Depends on:** #5663, #5660 (`message_broadcasts.channel`)

## Context

Staff work one inbox. The channel is a view on a contact, not a separate product surface.
`containers/WaGroups/GroupChatInterface` is an existing second chat surface alongside WhatsApp —
mirror its structure rather than inventing one.

### Backend
1. Expose `channel` on the message type in the GraphQL schema; add the `.gql` asset.
2. Surface the channel of the most recent message on each conversation so the list can badge it.
3. Add a channel argument to the per-contact message query. `Messages.list_conversation_messages/3`
   already filters on channel, and **that filter is the containment boundary**.
4. Add a channel filter to the conversation search in `searches.ex`.

### Frontend — `containers/Chat/ChatConversations`, `ChatMessages`, `graphql/queries/Search.ts`
5. Badge on the conversation list showing the latest message arrived on web.
6. Per-contact toggle between the web and WhatsApp views, driving the channel argument from step 3.
7. Web view shows whether the contact is currently connected; WhatsApp view keeps showing session information.

### Collection broadcasts pick a channel

Sending to a *collection* is not a separate screen — it is the ordinary composer with a collection
selected. `ChatMessages.tsx:141-144` swaps the mutation on `collectionId`, and the collection branch
calls `createAndSendMessageToGroup`, which lands in `Messages.create_and_send_message_to_group/3`
(`messages.ex:724`) and creates a `message_broadcasts` row with `type: "message"`, a `message_id` and
**no flow**. So `flows.channels` cannot gate it — the channel has to be an explicit input.

8. Add an optional `channel` to the `createAndSendMessageToGroup` input, **defaulting to `whatsapp`**
   so existing callers are unaffected. Persist it on `message_broadcasts.channel` at creation, since
   the broadcast worker drains the row in a later process and rebuilds its opts from it.
9. Dispatch each contact's message on that channel, so a web collection broadcast reaches connected
   widgets and a WhatsApp one is unchanged.
10. **Frontend:** a channel selector on the collection composer, visible only when the org has
    `web_channel_enabled`. Default WhatsApp.
11. `sendHsmMessageToGroup` (`message_types.ex:297`) is **API-only** — it appears nowhere in
    `glific-frontend`, and is documented in `api.docs/includes/_messages.md` with a Bruno entry. Leave
    its behaviour unchanged and WhatsApp-only: an HSM template has no meaning on the web channel, and
    it has external consumers we cannot see.

> **Reachability is caller-beware**, as with every other channel-initiating path (#5706). A web
> broadcast to a collection whose contacts have never opened the widget persists undelivered messages
> — there is no store-and-forward. Do not attempt to pre-filter the collection by reachability;
> nothing on master can answer that question until `contact_identities` lands (#5703).

## Acceptance criteria

- [ ] A contact with both histories shows two views and never mixes them.
- [ ] The badge is correct whichever channel the latest message arrived on.
- [ ] A contact with only WhatsApp history is unaffected.
- [ ] A collection broadcast sent on web reaches connected widgets, and the same broadcast with no
      channel selected behaves exactly as it does today.
- [ ] The channel selector is absent for an org without `web_channel_enabled`.

## Tests

Resolver tests for the channel filter; an **isolation test** that a WhatsApp message never appears
in a web-scoped query. Component tests for the toggle. **New e2e journey:** send from web, see it
badged in the inbox.

## Reviewer must verify personally

- [ ] **Read the isolation test.** A contact's WhatsApp history surfacing in the web view is a
      containment failure, not a formatting bug.
- [ ] Switch a contact with history on both channels back and forth and confirm nothing bleeds across.

rvignesh89:
Part of glific/glific#5713 and epic glific/glific#5659. The staff-side half of glific/glific#5717 (backend) — everything here is dark until `web_channel_enabled` is on for the organization, so this can merge independently.

Not stacked: `glific-frontend` has no earlier web-channel branch, so this targets `master`. It needs glific#5717 deployed to do anything.

## What staff get

**A channel selector above Contacts / Collections / Searches.** Above rather than beside, because the channel scopes what all three of those search — it is not a fourth thing to search. Picking Web writes `channel` into the conversation search filter, so the list shows only contacts who have a message on that channel.

**The same selector in the conversation header**, sharing one piece of state with the list: picking Web means "I am working the web channel", and it decides which transport your reply goes out on. The `createAndSendMessage` payload carries `channel` — omit it and the server defaults to WhatsApp, sending a browser visitor a message they never consented to, so that one line has a test of its own.

**Each channel's messages stay in their own view.** The contact list hides conversations that have no message on the selected channel, and a thread shows only that channel's messages — replying in the WhatsApp view under something the person said in the browser would otherwise answer on the wrong channel. A contact is one person across both channels; a thread is not.

**Presence instead of a session window.** A web conversation has no 24 hour window, so a countdown there would be a number that means nothing. The header shows online / offline from `contact.isWebOnline` instead, polled while the conversation is open (presence is ephemeral server state; there is no subscription for it, so a cached value would silently age).

**A composer that matches what the channel can do.** Everything in the inbox keys off `contactBspStatus`, which is a WhatsApp session concept: a browser-only contact sits at `NONE` and would otherwise be told "chat is unavailable, they aren't opted in to your number" — which on the web is simply untrue. On a web conversation the composer is always available, templates are not offered (there is no BSP to approve an HSM against), interactive templates, attachments, voice notes and emoji all work, and Start-a-flow is disabled with a tooltip saying why: the flow engine still never sees a web inbound message, so a flow started there would answer over WhatsApp.

**Its own colours.** Staff work both channels in one inbox, so the conversation surface says which one without being read: new `--app-color-channel-web` / `--app-color-surface-channel-web` tokens in `index.css`, indigo against WhatsApp's green. Both selector options are exactly equal width — "WhatsApp" is more than twice as long as "Web", and without that they read as two different controls.

## Deliberate calls worth arguing with

- **Both views filter at render, not in the query.** A filtered query reads a lazily-fetched result instead of the cache `ChatSubscription` writes into, so it would show the right conversations once and then stop updating live. Filtering what each view draws from an unfiltered, still-live cache keeps both channels correct *and* live, with no extra round trip. The cost is reach rather than correctness: a conversation whose loaded messages are all on the other channel is not listed until "load more" pulls in one that is. `searchFilter.channel` exists server-side for when that starts to matter.
- **A message with no channel counts as WhatsApp** — the server-side column default, and what anything cached before the field existed reads as. That is what keeps the flag-off path byte-identical.
- **One `channel` state, two selectors.** Switching in the header re-filters the list. That is the intended reading — you are working a channel, not a conversation.
- **Presence is polled, every 30s**, and reads as offline before the first response. Best-effort by design: a stale "Online" is worse than a cautious "Offline".

## Also here

- `SegmentedControl` gains an `equalWidth` prop rather than a new one-off component.
- `vitest.config.ts` now excludes `.claude/**`. Worktrees there are full checkouts of this repo, so every test file was being collected twice and the run reported roughly double the real count.

## Tests

New coverage for the selector, presence, the flag gate, the composer's web behaviour, the header's presence-vs-timer branch, and the channel in the send payload. The behavioural ones were mutation-checked — reverting the change makes them fail:

- restoring the `bspStatus === 'NONE'` opt-out message hides the web composer → 2 failures
- restoring the session timer on a web conversation → 1 failure

`src/containers/Chat`: 28 files, 285 tests, all passing.

Note for reviewers: `channel` is now selected on messages in `SEARCH_QUERY` and the three chat subscriptions, so every fixture that primes that cache had to carry it — a missing field makes a `cache-only` read return `undefined` and the inbox spins forever. That is what the mock churn in `src/mocks` is.

https://claude.ai/code/session_015JqHf9PqoGCW27fcNbQnnT

## Summary by CodeRabbit

- **New Features**
  - Added support for Web conversations alongside WhatsApp.
  - Added a channel selector for switching between WhatsApp and Web conversations.
  - Conversations and messages are filtered by the selected channel.
  - Web conversations display online/offline presence indicators.
  - Added Web-specific message sending and interactive message support.
  - Added channel-aware layouts and responsive styling.

- **Bug Fixes**
  - Web conversations no longer show WhatsApp opt-in warnings or session timers.

- **Documentation**
  - Added English and Hindi translations for Web presence and channel-related messages.


src/mocks/Ticket.tsx
```diff
@@ -58,6 +58,7 @@ export const ticketListQuery = {
           },
           id: '1',
           messageNumber: 23,
+          channel: 'WHATSAPP',
```
rvignesh89, line 61: This should be smallcaps

AmishaBisht: Cross checked this, the API returns uppercase

AmishaBisht (approved):

PR #4176, opened 2026-09-09
fix: disable Run evaluation when the version has not been built

### LINKED ISSUE #5728: Block AI Evaluation creation for assistants not synced to Kaapi
**Priority:** P2
**Type:** Bug (frontend + backend)
**Component:** AI Evaluations / Assistants

## Observed

```
%{status: 422, body: %{error: "Validation failed", errors: [
  %{message: "Field required", field: "config_id"},
  %{message: "Field required", field: "config_version"}]}}
```

## Root cause

`create_evaluation` builds the Kaapi payload from the assistant's Kaapi identity without checking it exists — `lib/glific_web/resolvers/ai_evaluations.ex:438-443`:

```elixir
config_id: config_version.assistant.kaapi_uuid,
config_version: config_version.kaapi_version_number,
```

For an assistant that was never synced to Kaapi both are `nil` (see the explicit `create_kaapi_config_version(%{kaapi_uuid: nil}, ...)` no-op at `lib/glific/assistants.ex:773`), so we POST `config_id: null, config_version: null` and Kaapi rejects it with a 422 that reaches the user as a raw validation dump. Reproduced by creating an eval against an assistant with no UUID.

## Scope of fix

1. **Backend guard (do this regardless of the UI change)** — reject before the HTTP call and return the message we already use elsewhere. `Assistants.fetch_kaapi_config/3` (`lib/glific/assistants.ex:1236-1257`) is the existing precedent; it guards `not is_nil(kaapi_uuid)` / `not is_nil(kaapi_version_number)` and returns `"Assistant is not available on Kaapi yet"` / `"Selected assistant version is not available on Kaapi yet"`. Mirror that in the `with` chain of `create_evaluation`, before `run_kaapi_evaluation/3`.
2. **Frontend** — disable the Create/Run AI Evaluation action when the selected assistant or config version has no Kaapi identity, with a tooltip explaining why. Both fields are already exposed on the GraphQL schema, so no API change is needed: `kaapi_uuid` on `:assistant` (`lib/glific_web/schema/assistant_types.ex:61`) and `kaapi_version_number` on the config-version object (`:193`).

## Acceptance criteria

- Eval creation for an assistant with `kaapi_uuid == nil` returns a human-readable error, never a Kaapi 422 body, and makes no outbound Kaapi call.
- Same for a config version with `kaapi_version_number == nil`.
- The UI button is disabled (not merely erroring on submit) for such assistants, with a tooltip.
- Resolver test covering both nil cases; assert no Tesla request is issued.

**Out of scope:** why an assistant ends up without a UUID (retry/backfill of Kaapi sync) — worth a separate ticket if it's recurring in prod.

Ayush8923:
Closes https://github.com/glific/glific/issues/5728

## Summary
- **Run evaluation** stayed enabled on a version the server never finished building, so starting a run on it was a dead end. It now behaves like **Go Live**, disabled, with a tooltip on hover saying why.
    - **failed version:** "This version failed to build, so it cannot be evaluated. Save a new version."
    - **version still building:** "This version is still being prepared. It can be evaluated once the server has finished building it."
    - **run already going:** unchanged, keeps its existing message
- The AI summary now renders its markdown
-  A three-digit row number no longer wraps in the results table

### DIFF
--- a/src/containers/AIEvaluation/AssistantDetail/Tabs/KnowledgeBase/KnowledgeBase.tsx
+++ b/src/containers/AIEvaluation/AssistantDetail/Tabs/KnowledgeBase/KnowledgeBase.tsx
@@ -114,7 +114,7 @@ export const KnowledgeBase = ({
         return;
       }
 
-      downloadFile(signedUrl, data?.getFile?.filename || file.filename);
+      downloadFile(signedUrl, data?.getFile?.filename || file.filename, '_self');
     } catch (error: unknown) {
       setErrorMessage(error);
     } finally {
--- a/cypress/e2e/aiAssistant/AssistantEvaluation.spec.ts
+++ b/cypress/e2e/aiAssistant/AssistantEvaluation.spec.ts
@@ -1,7 +1,9 @@
 import { BELOW_STICKY_HEADER } from '../../utils/assistant-flow';
 import {
+  BUILDING_VERSIONS,
   EMPTY_SCORES,
   FAILED_RUN,
+  FAILED_VERSIONS,
   GOLDEN_SETS,
   LIVE_VERSION_ID,
   RUNNING_RUN,
@@ -10,7 +12,12 @@ import {
   UNSCORED_SCORES,
   WEAK_SCORES,
 } from '../../utils/assistant-fixtures';
-import { openAssistant, openAssistantPage, openTab } from '../../utils/assistant-detail';
+import {
+  openAssistant,
+  openAssistantPage,
+  openTab,
+  selectVersion,
+} from '../../utils/assistant-detail';
 
 describe('the Golden Q&A Evaluation tab of a saved assistant', () => {
   beforeEach(() => {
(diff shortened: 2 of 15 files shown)


src/containers/AIEvaluation/AssistantDetail/Tabs/KnowledgeBase/KnowledgeBase.tsx
```diff
@@ -114,7 +114,7 @@ export const KnowledgeBase = ({
         return;
       }
 
-      downloadFile(signedUrl, data?.getFile?.filename || file.filename);
+      downloadFile(signedUrl, data?.getFile?.filename || file.filename, '_self');
```
akanshaaa19, line 117: why are we replacing this with _self

akanshaaa19: it will open in the same tab- do we want that?

akanshaaa19 (approved):
looks fine

src/containers/AIEvaluation/AssistantDetail/Tabs/KnowledgeBase/KnowledgeBase.tsx
akanshaaa19: i suggest to move it back to _target

Ayush8923: since the file will now be downloaded directly instead of being viewed, I don’t think we need to open a new tab for the download. Otherwise, a new tab will open, the file will download, and then we will have to go back to the previous tab. It would be better to trigger the download directly in the same tab.

